PR #5, opened 2013-09-10
Export CMake targets

plfiorini:
Improve CMake configuration files by exporting targets and put version
information in a dedicated file.

Library is exported with its full path thus making it usable by other
projects. Relative paths are used to make it relocatable.

For more information see:
  http://www.cmake.org/Wiki/CMake/Tutorials/Packaging
  http://www.cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets
  http://www.cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file

While we are at it, rename library to QtXdg for Qt 4 builds and Qt5Xdg
for Qt 5 builds.


jleclanche:
@luis-pereira What happened with this?

jleclanche:
@luis-pereira what do we do with this?

luis-pereira:
Will be handled after the release.

luis-pereira:
Done by https://github.com/lxde/libqtxdg/commit/b656ffa55f798762150a4e546779b61836a90d7f
@plfiorini Thanks, you pointed the way.

PR #36, opened 2014-11-23
add fallback for unthemed app icons

dbluelle:
some apps don't have their icons in "hicolor" dir
and just put an icon in the base directory (e.g the official Qt maintenance application)
If I understand the xdg specs correctly, this is allowed as a last fallback if no matching icon is found elsewhere

### DIFF
--- a/qiconfix/qiconloader.cpp
+++ b/qiconfix/qiconloader.cpp
@@ -384,6 +384,34 @@ QThemeIconEntries QIconLoader::findIconHelper(const QString &themeName,
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


jleclanche:
@luis-pereira ping on this, can we land before 0.9?

luis-pereira:
@jleclanche @dbluelle Reviewing it

luis-pereira:
Thanks @dbluelle landed on https://github.com/lxde/libqtxdg/commit/ff25cc42c0c7023f9cf962491ceab5912100bf8f

PR #41, opened 2015-02-04
Use gcc visibility attribute and Q_DECL_EXPORT macro to avoid exporting internal symbols.

PCMan:
This greatly reduced the size of dynamic symbol table and might potentially speed up loading.
Also, turn off exception handling to save more since we don't use it.
For non-gcc compilers, this change does nothing.
@luis-pereira please see if this is OK for you. :-)
Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,12 @@ set (MINOR_VERSION 0)
 set (PATCH_VERSION 0)
 set(QTXDG_VERSION_STRING ${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION})
 
-add_definitions(-Wall)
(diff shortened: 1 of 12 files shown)


luis-pereira:
@PCMan I get this warning:
> warning: ‘visibility’ attribute ignored [-Wattributes]

from:
```static const QTXDG_API QString userDirectoryString[8]```

gcc is right, ```static const``` implies hidden visibility. QTXDG_API should be removed in this case. Other than that, it's good to go :+1:

PCMan:
OK, fixed.
BTW, will it be better to move that array in a *.cpp file instead?
Putting a static variable in a public header looks a little bit weird.

luis-pereira:
@PCMan good point. I will do it.
I will make this patch work for Clang also.

PR #46, opened 2015-03-02
Fix missing slash in paths in xdgDirList

paulolieuthier:
Fixes lxde/lxqt#518.

### DIFF
--- a/xdgdirs.cpp
+++ b/xdgdirs.cpp
@@ -98,7 +98,11 @@ QStringList xdgDirList(const QString &envVar, const QString &postfix)
     for (QStringList::Iterator i=dirs.begin(); i!=dirs.end(); ++i)
     {
         fixBashShortcuts((*i));
-        *i += postfix;
+        if (!(*i).isEmpty()) {
+            if (!(*i).endsWith(QLatin1Char('/')))
+                *i += "/";
+            *i += postfix;
+        }
     }
     return dirs;
 }


jleclanche:
hold

palinek:
Just looked quickly at the ```xdgdirs.cpp``` and there are path fiddlings by hand all around. Is there a reason to not do it in portable way... e.g. using ```boost::filesystem::path``` (http://www.boost.org/doc/libs/1_57_0/libs/filesystem/doc/reference.html#class-path)

jleclanche:
@palinek I was saying on IRC that we should be using QDir::cleanPath() or some such.

We don't depend on boost right now btw, and we won't.

paulolieuthier:
The hard-coded paths are standard paths according to the xdg spec. No problem with that.  I think the patch is appropriate because `XdgDirs::dataDirs()` must return valid complete paths. Using `QDir::cleanPath()` would remove the ending slash of the paths that already have it and we would have to append the slash anyway. I'm just checking if there is an ending slash and appending one if there is not.

luis-pereira:
```QDir::cleanPath()``` doesn't solve this issue. It's ending slash behaviour is not specified so we must not rely on that. I'm not against using ```QDir::cleanPath()``` tough. 
@paulolieuthier may I push onto this branch ?

jleclanche:
What about `QDir::canonicalPath`?

paulolieuthier:
@luis-pereira I tested `QDir::cleanPath` and it removes the ending slash. Please, feel free to work on this branch.

luis-pereira:
@paulolieuthier I know that it does that. But it's a not documented behaviour.

jleclanche:
I'd rather rely on undocumented but proven behaviour than an internal hack :)

paulolieuthier:
I'm in favour of whatever you guys decide, I just want to see the bug fixed :smile:. But it doesn't look like a hack to me, we're just adding a slash if there is not one.

luis-pereira:
I wouldn't also call this patch a hack. It appends an trailing / ```QDir::cleanPath()``` does the opposite in an undocumented way. QFileSystemWatcher also worked fine in an undocumented way, until it got changed. 
@paulolieuthier @jleclanche Updated.

paulolieuthier:
Much better. GTM

jleclanche:
Alright, gtm. Squash and ship please.

PR #47, opened 2015-03-03
Qt5: Use Qt5 private headers

luis-pereira:
Also installs our private header (qiconloader_p.h). This way QIconLoader
can be accessed by user code.
By accessing Qt5 private headers we can access functionality that was
previously not accessible (QGuiApplicationPrivate).

We also install our own private header. This way the user can access our
QIconLoader if necessary. Accessing QIconLoader is for special uses cases,
so we "hide" it's header under ${QTXDG_PRIVATE_INCLUDE_DIR}.


luis-pereira:
@paulolieuthier @jleclanche @agaida Pls test before going to master.

paulolieuthier:
GTM.

Issue #50, opened 2015-03-12
XdgDesktopFile: Not executing Type=Application with pkexec on the Exec line


luis-pereira:
The use of ```QProcess::startDetached()``` in ```XdgDesktopFile::startDetached()``` makes pkexec exit with "Refusing to render service to dead parents." due to double forking. http://cgit.freedesktop.org/polkit/tree/src/programs/pkexec.c#n730

References:
https://issues.openmandriva.org/show_bug.cgi?id=1128#c3

palinek:
I don't know how the ```XdgAction``` or ```XdgDesktopFile``` are used (if the calling/executin application can finish etc..), but I had a quick look at this and the solution for pkexec can be something like this:
```c++
QProcess* p = new QProcess;
p->setStandardInputFile(QProcess::nullDevice());
//XXX: we probably should redirect stdout/stderr to our stdout/stderr
p->setStandardOutputFile(QProcess::nullDevice());
p->setStandardErrorFile(QProcess::nullDevice());
connect(p, &QProcess::finished, p, &QObject::deleteLater);
p->start(cmd, args);
return p->waitForStarted();
```

palinek:
Just played with this one and IMO there is no "easy & nice" solution. Following is possible:

1. use the above, but the newly spawned process remains a direct child of the parent process (lxqt-runner, lxqt-panel ... etc) => if the parent process ends, the child exits also (gets a SIGHUP)

2. if we want to completely detach from parent process we must implement our wrapper which would completely detach from parent (two ```fork```s, ```setsid```... etc), after that spawns required child (```exec```) and just wait for it's exit => all created process's will have dumb running parent process

3. use the above only as a workaround for ```pkexec``` (or any other known in the future)

palinek:
Sorry for misleading with the above... all that is true only in case the parent process has a tty attached. That's not the case of lxqt components run by DM.
Just tried the above solution with lxqt-panel (with no tty) => pkexec (in .desktop) works and application spawned by lxqt-parent don't exit upon killing lxqt-panel.

... will make PR

@paulolieuthier the ```QProcess::startDetached``` is also used in lxqt-runner so running "pkexec <application>" don't work either there. Maybe we should use this solution also there...

PR #51, opened 2015-03-13
desktopfile: run new process as a direct child

palinek:
fixes lxde/libqtxdg#50

### DIFF
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -394,7 +394,17 @@ bool XdgDesktopFileData::startApplicationDetached(const XdgDesktopFile *q, const
     }
 
     QString cmd = args.takeFirst();
-    return QProcess::startDetached(cmd, args);
+    QScopedPointer<QProcess> p(new QProcess);
+    p->setStandardInputFile(QProcess::nullDevice());
+    p->setProcessChannelMode(QProcess::ForwardedChannels);
+    p->start(cmd, args);
+    bool started = p->waitForStarted();
+    if (started)
+    {
+        QProcess* proc = p.take(); //release the pointer(will be selfdestroyed upon finish)
+        QObject::connect(proc, SIGNAL(finished(int, QProcess::ExitStatus)), proc, SLOT(deleteLater()));
+    }
+    return started;
 }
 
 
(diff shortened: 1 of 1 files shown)


luis-pereira:
@palinek Great work. But this means that every app won't be detached. IMO we have to make a compromise here and create a list of non detach execs,

palinek:
@luis-pereira why is detaching so critical?

PCMan:
If not detached, a new process will become a child of the process which
calls the API.
Then when the parent process, for example lxqt-panel, is killed, all of the
apps launched by it will be killed as well, which is not we want.
So, the new processes should be detached.
2015/3/18 下午3:01 於 "palinek" <notifications@github.com> 寫道：

palinek:
@PCMan that's what I thought also... but that's truth only in case the parent is a session leader and has attached terminal .. see my "selfdiscussion" in lxde/libqtxdg#50

luis-pereira:
@palinek You are totally right in your self discussion.
The case where the parent has an attached terminal is important for developers (at least). Most of the times, during debugging and testing, developers launch the panel (for example) from a terminal. In that case killing the terminal will also kill any panel launched apps. I'm not looking forward to that.

@palinek @PCMan  Here's a draft of my proposal, based on @palinek work: https://github.com/lxde/libqtxdg/tree/add-nondetach-exec-list

palinek:
@luis-pereira the decision is up to you. But your solution does not handle e.g `Exec=/usr/bin/pkexec foo`, the check should be probably done only on the basename

luis-pereira:
@palinek It does indeed handle it. ```QString::contains()``` does it. If we check only the basename (command) we miss something like this: ```Exec=sh -c "/usr/bin/pkexec gparted"```

palinek:
@luis-pereira i see (i was blind before)...

paulolieuthier:
@palinek pull @luis-pereira's commits to your branch so it can be merged.

palinek:
@paulolieuthier done

luis-pereira:
@palinek Pls pull it again, sorry. I added some minor documentation and wrote an commit message.

palinek:
@luis-pereira done... should this be landed?

luis-pereira:
@palinek Yes, GTM.

PR #55, opened 2015-03-28
Fixes XdgDesktop::isShow() caching mechanism

luis-pereira:
The cache mechanism was totally broken. It stored the last mIsShow value.
And that's it. It didn't take into account the environment.
It didn't work with straightforward code like this:
for (int i = 0; i < N; ++i)
{
    if (file->isShown(envs.at(i)))
    {
        show = true;
        break;
    }
}

Closes lxde/lxqt#563

### DIFF
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -298,7 +298,7 @@ public:
     QString mFileName;
     bool mIsValid;
     mutable bool mValidIsChecked;
-    mutable Triple mIsShow;
+    mutable QHash<QString, bool> mIsShow;
     QMap<QString, QVariant> mItems;
 
     XdgDesktopFile::Type mType;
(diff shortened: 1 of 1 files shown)


palinek:
@luis-pereira that's what i suggested...not?

luis-pereira:
@palinek With a minor difference: bool instead of Triple

palinek:
I see... don't need the undef state...

palinek:
@luis-pereira why did you assign me? I don't know deeply about the code & usage of libqtxdg... therefore didn't make the changes alone and asked you.

xdgdesktopfile.cpp
```diff
@@ -298,7 +298,7 @@ class XdgDesktopFileData: public QSharedData {
     QString mFileName;
     bool mIsValid;
     mutable bool mValidIsChecked;
-    mutable Triple mIsShow;
+    mutable QHash<QString, bool> mIsShow;
```
palinek: `enum Triple` is now used nowhere

luis-pereira:
@palinek So that you can learn the code. :smile:  You did found the problem and proposed a correct solution.
I'm thinking about the bus factor here. It's good that other active developers know the code.

I was committed in the moment I got the message about the Triple.

palinek:
Landed in d2c9635
fixes lxde/lxqt#563

Issue #54, opened 2015-03-28
Support  “-” to separate levels of specificity in icon names


luis-pereira:
As described in http://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#guidelines

PR #57, opened 2015-04-02
Backports Fixed icons lookup in QIcon::fromTheme

luis-pereira:
We already supported multile directories. But we didn't support dashes to
separate levels of specificity in icon names.
Closes lxde/libqtxdg#54.

QIconloader upstream commit:
https://qt.gitorious.org/qt/qtbase/commit/a40d390a3bbc1919843000b427a168180b83a344

Upstream commit message:
This commit fixes incorrect logic of icons' lookup if there are
fallbacks or more than one theme's directory.

According to Icon Theme Specification, Directory Layout section, theme
can be spread across several base directories by having subdirectories
of the same name. This makes possible to extend system themes by
application-specific icons without making of collisions with other
applications.

According to Icon Naming Specification, Icon Naming Guidelines section,
icon name may contain dashes to separate levels of specificity in icon
names. This makes possible to set in application very specific icon
which may be not in every theme. So it can fallback to less specific one.

### DIFF
--- a/qiconfix/qiconloader.cpp
+++ b/qiconfix/qiconloader.cpp
@@ -176,7 +176,7 @@ QIconTheme::QIconTheme(const QString &themeName)
         QString themeDir = iconDir.path() + QLatin1Char('/') + themeName;
         themeIndex.setFileName(themeDir + QLatin1String("/index.theme"));
         if (themeIndex.exists()) {
-            m_contentDir = themeDir;
+            m_contentDirs << themeDir;
             m_valid = true;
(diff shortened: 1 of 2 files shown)


luis-pereira:
Before this patch:
screen-2015-04-02-16-11-22

After this patch:
screen-2015-04-02-16-10-11

luis-pereira:
Landed at lxde/libqtxdg@ab0f65cd6b5b695d24c5020c1b31104a989728d4

PR #64, opened 2015-07-13
CMake: Adds Runtime install COMPONENT

luis-pereira:
There are two install components:
    * Devel Contains all the development files;
    * Runtime Contains all the runtime needed files.

It's possible to install:
  - Everything: make install OR cmake -P cmake_install.cmake
  - Development files only: cmake -DCOMPONENT=Devel -P cmake_install.cmake
  - Runtime files only: cmake -DCOMPONENT=Runtime -P cmake_install.cmake

This COMPONENT's can also be used with CPack.


luis-pereira:
I'm planning on replicate this everywhere.... So the time to speak is now.

jleclanche:
I don't see the problem with this.

PR #68, opened 2015-12-22
use doxygen and qhelpgenerator to make qch file for Assistant/Qt Creator

ec1oud:
"make qch" works now for me, and then I can get help in Creator on classes in this module.  I'm not a cmake expert, so not sure of the best way to ensure that qhelpgenerator really exists.


luis-pereira:
@ec1oud This patch has several issues:

- Lack of an BUILD_DOCUMENTATION option
- We can't assume that qhelpgenerator exists
- To get the documentation one as to call it explicitly: make doc or make qch
- make doc or make qch don't work. There's no doxygen config file

Sorry for the late reponse.

agaida:
no action in two years - closing

PR #69, opened 2016-01-18
xdgdesktopfile: handle 'Path' entry for application's working directory

schnitzeltony:
it is an optional enty of the specification [1]

http://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html

Signed-off-by: Andreas Müller <schnitzeltony@googlemail.com>


paulolieuthier:
GTM.

palinek:
GTM

PR #74, opened 2016-02-26
XdgUserDirs: Return a fallback when the path is empty

luis-pereira:
If the path to the folder is empty we return the default value.
It's what xdg-user-dirs also do.


paulolieuthier:
GTM.

jleclanche:
gtm

PR #76, opened 2016-03-02
Adds XdgDesktopFile::mimeTypes() member

luis-pereira:
A convenience, but handy, member.

### DIFF
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -68,6 +68,7 @@ static const char onlyShowInKey[] = "OnlyShowIn";
 static const char notShowInKey[] = "NotShowIn";
 static const char categoriesKey[] = "Categories";
 static const char extendPrefixKey[] = "X-";
+static QLatin1String mimeTypeKey("MimeType");
 
 // Helper functions prototypes
 bool checkTryExec(const QString& progName);
@@ -767,6 +768,33 @@ QIcon const XdgDesktopFile::icon(const QIcon& fallback) const
 QString const XdgDesktopFile::iconName() const
 {
     return value("Icon").toString();
+}
+
+
+QStringList XdgDesktopFile::mimeTypes() const
+{
+    if (contains(mimeTypeKey))
+    {
+        QStringList s = value(mimeTypeKey).toString().split(QLatin1Char(';'));
+        if (s.isEmpty())
+        {
+            return s;
+        }
+        else
+        {
+            // Remove empty strings
+            QMutableListIterator<QString> it(s);
+            while (it.hasNext())
+            {
+                const QString dir = it.next();
+                if (dir.isEmpty())
+                    it.remove();
+            }
+        }
+        return s;
+    }
+
+    return QStringList();
 }
 
 
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -773,28 +773,7 @@ QString const XdgDesktopFile::iconName() const
 
 QStringList XdgDesktopFile::mimeTypes() const
 {
-    if (contains(mimeTypeKey))
-    {
-        QStringList s = value(mimeTypeKey).toString().split(QLatin1Char(';'));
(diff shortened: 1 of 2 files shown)


paulolieuthier:
GTM.

xdgdesktopfile.cpp
```diff
@@ -770,6 +771,33 @@ QString const XdgDesktopFile::iconName() const
...
+    {
+        QStringList s = value(mimeTypeKey).toString().split(QLatin1Char(';'));
+        if (s.isEmpty())
+        {
+            return s;
+        }
+        else
+        {
+            // Remove empty strings
+            QMutableListIterator<QString> it(s);
```
palinek: I was always curious about this Qt iterating style.... Especialy here, is there any benefit over the stl-like manipulation?
E.g.:
- pre C++11
```diff
diff --git a/xdgdesktopfile.cpp b/xdgdesktopfile.cpp
index b27070c..f379aaf 100644
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -33,6 +33,7 @@
 
 #include <stdlib.h>
 #include <unistd.h>
+#include <functional>
 
 #include <QDebug>
 #include <QDBusInterface>
@@ -783,13 +784,7 @@ QStringList XdgDesktopFile::mimeTypes() const
         else
         {
             // Remove empty strings
-            QMutableListIterator<QString> it(s);
-            while (it.hasNext())
-            {
-                const QString dir = it.next();
-                if (dir.isEmpty())
-                    it.remove();
-            }
+            s.erase(std::remove_if(s.begin(), s.end(), std::mem_fun_ref(&QString::isEmpty)), s.end());
         }
         return s;
     }
```
- post C++11
```diff
diff --git a/xdgdesktopfile.cpp b/xdgdesktopfile.cpp
index b27070c..9d1e759 100644
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -33,6 +33,7 @@
 
 #include <stdlib.h>
 #include <unistd.h>
+#include <functional>
 
 #include <QDebug>
 #include <QDBusInterface>
@@ -783,13 +784,7 @@ QStringList XdgDesktopFile::mimeTypes() const
         else
         {
             // Remove empty strings
-            QMutableListIterator<QString> it(s);
-            while (it.hasNext())
-            {
-                const QString dir = it.next();
-                if (dir.isEmpty())
-                    it.remove();
-            }
+            s.erase(std::remove_if(std::begin(s), std::end(s), std::bind(&QString::isEmpty, std::placeholders::_1)), std::end(s));
         }
         return s;
     }
```

palinek: This raises another question: why aren't we using c++11 in libqtxdg at all?

luis-pereira:
@palinek We aren't using C++11 because..... no reason actually.
I'm not aware of any advantage or disadvantage in this Qt iterating style over std. I just used what I'm really used to. It may sound strange but I recognized the iterator pattern quicker than std C++11 one. It's obviously an matter of training.

jleclanche:
We're using C++11 in the rest of LXQt, so there's no opposition I know of to just use it here too.

palinek:
@luis-pereira, please, consider the @1a64f06. I believe it will do the same, but in simpler way.

palinek:
@luis-pereira I've just "force-pushed" the these two commits squashed together over your merge. To avoid mess in the git log/history.

luis-pereira:
That's fine :)

On Mon, Mar 14, 2016 at 4:05 PM, palinek <notifications@github.com> wrote:

> @luis-pereira <https://github.com/luis-pereira> I've just "force-pushed"
> the these two commits squashed together over your merge. To avoid mess in
> the git log/history.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lxde/libqtxdg/pull/76#issuecomment-196386248>.
>



-- 
        Luís Pereira

PR #80, opened 2016-03-18
XdgDesktopFile: Adds Exec fallback when DBusActivatable fails

luis-pereira:
We fallback to use the Exec key when the DBusActivatable fails.
Closes lxde/lxqt#967.

### DIFF
--- a/xdgdesktopfile.cpp
+++ b/xdgdesktopfile.cpp
@@ -355,6 +355,27 @@ bool XdgDesktopFileData::startApplicationDetached(const XdgDesktopFile *q, const
 {
     //DBusActivatable handling
     if (q->value(QLatin1String("DBusActivatable"), false).toBool()) {
+        /* WARNING: We fallback to use Exec when the DBusActivatable fails.
+         *
+         * This is a violation of the standard and we know it!
+         *
+         * From the Standard:
+         * DBusActivatable	A boolean value specifying if D-Bus activation is
+         * supported for this application. If this key is missing, the default
+         * value is false. If the value is true then implementations should
+         * ignore the Exec key and send a D-Bus message to launch the
+         * application. See D-Bus Activation for more information on how this
+         * works. Applications should still include Exec= lines in their desktop
+         * files for compatibility with implementations that do not understand
+         * the DBusActivatable key.
+         *
+         * So, why are we doing it ? In the benefit of user experience.
+         * We first ignore the Exec line and in use the D-Bus to lauch the
+         * application. But if it fails, we try the Exec method.
+         *
+         * We consider that this violation is more acceptable than an failure
+         * in launching an application.
+         */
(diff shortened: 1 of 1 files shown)


jleclanche:
gtm

palinek:
I was also considering this, but it violates the specification "If the value is true then implementations should ignore the Exec key and send a D-Bus message to launch the application."

luis-pereira:
@palinek But the specification also says:

>  Even if DBusActivatable is true, Exec should be specified for compatibility with implementations that do not understand DBusActivatable.

palinek:
> that do not understand DBusActivatable

But we do.

luis-pereira:
Yes. But how do libqtxdg determine if a given environment understands or not  `DBusActivatable` ?

palinek:
> Yes. But how do libqtxdg determine if a given environment understands or not DBusActivatable ?

Libqtxdg is the environment/implementation here, not?

luis-pereira:
Sure. Not exactly what I meant.
What I meant is that even the standard guys do not entirely follow the standard.
When `DBusActivatable` was introduced at least desktop-file-validate should got update to warn about the presence of `-` in the filename.

We strictly follow the standard and get a fail (user viewpoint).

palinek:
> We strictly follow the standard and get a fail (user viewpoint).

From user point of view it is definitely better to try the `Exec` as a fallback if `DBus` fails... there is no doubt.

Let's do it. But IMO it has to be documented somehow (comment in code, readme?), that we are violating the standard for a good reason.

luis-pereira:
@palinek I do agree with the documentation). Will write it.

luis-pereira:
@palinek Documented at https://github.com/lxde/libqtxdg/pull/80/commits/d019654dc3d83f7986f8272c5ae39650364b532d. Take a look.

palinek:
It's a bit too verbose... But the fallback executed only for whatever "fail" reason (no dbus at all, not properly configured the dbus service for the applicaion...), not only the invalid name.

luis-pereira:
@palinek It's verbose, but IMO it deserves a very good explanation.
The fallback is now only executed when there is an invalid dbus object path.

palinek:
> The fallback is now only executed when there is an invalid dbus object path.

Hm... why? From end user POI it doesn't matter what triggered the error. With my previous note I proposed a change in comment/doc, not the change of the previous logic.

luis-pereira:
@palinek I inferred from your word that you only agreed in doing it in that special casw

palinek:
> that you only agreed in doing it in that special case

Sorry for confusing you :smile:

luis-pereira:
@palinek No problem. Updated.

paulolieuthier:
GTM.

palinek:
GTM

PR #82, opened 2016-04-20
Minor performance

luis-pereira:


xdgdesktopfile.cpp
```diff
@@ -935,7 +935,7 @@ QString expandEnvVariables(const QString str)
 QStringList expandEnvVariables(const QStringList strs)
 {
     QStringList res;
-    foreach(QString s, strs)
+    foreach(const QString s, strs)
```
paulolieuthier: Missed `&` here?

luis-pereira: yup

paulolieuthier:
GTM.

palinek:
GTM

PR #83, opened 2016-05-05
Improve foreach iteraror performance

luis-pereira:
We use const references where appropriate. Also made the container constant
when appropriate.


palinek:
GTM

PR #88, opened 2016-05-27
Fix several issues discovered with analyze tools

luis-pereira:


luis-pereira:
Analyze & debug in progress

qtxdg/xdgdesktopfile.cpp
```diff
@@ -1480,8 +1480,16 @@ void XdgDesktopFileCache::initialize(const QString& dirName)
 XdgDesktopFile* XdgDesktopFileCache::load(const QString& fileName)
 {
     XdgDesktopFile* desktopFile = new XdgDesktopFile();
-    desktopFile->load(fileName);
-    return desktopFile;
+
+    Q_CHECK_PTR(desktopFile);
+    if (desktopFile && desktopFile->load(fileName))
+    {
+        desktopFile->load(fileName);
```
paulolieuthier: Twice?

luis-pereira: ooooppps, didn't commit. It's time to rest :wink:

qtxdg/xdgdesktopfile.cpp
```diff
@@ -836,9 +836,11 @@ bool XdgDesktopFile::startDetached(const QStringList& urls) const
     {
     case ApplicationType:
         return d->startApplicationDetached(this, urls);
+        break;
```
jubalh: if one makes `return` the `break` doesnt have much effect, right?

luis-pereira: t@jubalh Sure. But the return might be replaced by something else and adding the `break` might be forgot.
Also, it confuses analyzers.

luis-pereira:
It has been working without glitches. It needs to go mainstream.

PR #90, opened 2016-05-30
qtxdg: Get the HOME environment variable just one time

luis-pereira:
Avoids unnecessary string conversions.


paulolieuthier:
GTM.

palinek:
GTM

PR #91, opened 2016-06-03
XdgDesktopFileData: Explicitly initialize members

luis-pereira:
Explicitly initialize all members.

### DIFF
--- a/qtxdg/xdgdesktopfile.cpp
+++ b/qtxdg/xdgdesktopfile.cpp
@@ -294,8 +294,12 @@ public:
 
 
 XdgDesktopFileData::XdgDesktopFileData():
+    mFileName(),
     mIsValid(false),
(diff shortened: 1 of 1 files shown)


paulolieuthier:
Explicitly initialization of stack-allocated objects that don't need parameters? Weird... Anyway, GTM.

palinek:
I also think, that default constuctible objects don't need to be explicitly mentioned.

luis-pereira:
They strictly don't need, sure.
The starting point for this was `mType` not being initialized in the constructor or any of `XdgDesktopFileData` members. And it really should, in order to avoid future issues when updating the code.
So I ended up, just initializing everything explicitly to make it clear, that when adding new stuff to `XdgDesktopData` one should also think about initialization.
That's it.

agaida:
GTM

PR #94, opened 2016-06-16
build: Create separate Qt5XdgIconLoader target

palinek:
When trying to compile lxqt-runner by use of the binary packages from debian's lxqt-snapshots I was facing following problem:
```
CMake Error in CMakeLists.txt:
  Imported target "lxqt" includes non-existent path

    "/usr/include/qt5xdgiconloader"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

So I've realized that there are 2 problems with the newly split libqt5xdg package:
- the package `libqt5xdg-dev` doesn't depend on `libqt5xdgiconloader-dev` (downstream @agaida)
- the qt5xdgiconloader can't be included in "pure cmake" projects -> the qt5xdgiconloader*.cmake files are not generated (upstream)

This updates the build to generate qt5xdgiconloader*.cmake files and gives the dependency of Qt5XdgIconLoader to Qt5Xdg.

@luis-pereira Please, review this carefully as my skills in this area are realy poor.


luis-pereira:
@palinek I will look into it.

luis-pereira:
@palinek
A `config.cmake` file and a target are not the same thing. IMO, we should have one `config.cmake` for the all package. We can have separated target files, but I really don't see the point.
This seems to me more a downstream packaging issue. Just made an "demonstration". Check it [here](https://gist.github.com/luis-pereira/23c33728bdc9b4723d49ff77e66a3cb2). Everything is working Ok.

palinek:
> A config.cmake file and a target are not the same thing. IMO, we should have one config.cmake for the all package. We can have separated target files, but I really don't see the point.

If we don't have separate `config.cmake` files, then the separation of `*-dev` packages (as it is now done in debian) is not possible.

My thoughts: if the Qt5XdgIconLoader is a separate library (one doesn't need to use anyghing else from Qt5Xdg), why can't the Qt5XdgIconLoader be a separate dev module/package?

luis-pereira:
@palinek I do get that the separation of the *-dev packages is not possible. But why were they split in the first place ?
It's a separate library on the same package. If we create a package for each library, maybe we should also put it in a separate repository.

palinek:
> I do get that the separation of the *-dev packages is not possible. But why were they split in the first place ?
It's a separate library on the same package. If we create a package for each library, maybe we should also put it in a separate repository.

@agaida ?

agaida:
If there are no dramatic changes i'm fine with the situation right now - build the needed parts from one source, split the compiled things into separated packages, be done with :)
https://anonscm.debian.org/cgit/pkg-lxqt/libqtxdg.git/tree/debian?h=lxqt-snapshots/latest

agaida:
I really don't know if the target thing is really needed - but
a) it doesn't hurt
b) it can help to make things more clear for other people

so - GTM

luis-pereira:
@agaida A Target is always needed. The discussion is about a separate target file and a separate package. What the PR actually implements is a new Qt5XdgIconloader package.
I really don't understand the migration an whining stuff, but I'm not a packager.

From my POV it was good the way it was. This PR is technically correct, anyway.
Landed at  a8f82dff82b3dbada583a99a2ca9a4f28a55205a.

agaida:
@luis-pereira - sure, a target is needed. And it look better now, but would have worked without this change. It's just cleaner. Lintian is a set of scripts that check the packaging bugs and so on and it is really good at it - prevent lintian whining means that your packages are formal correct.

PR #100, opened 2016-08-01
Adds Qt5Svg explicitly to the required packages

luis-pereira:
It was an implicit dependency as XdgIconLoader always support svg icons.
Let's make it explicit.


paulolieuthier:
GTM.

palinek:
Gtm

PR #104, opened 2016-09-20
test: Fixes false positive in tst_xdgdesktopfile::testReadLocalized()

luis-pereira:
LC_MESSAGES is the first in the hierarchy. Using LANG can lead to false
positives.

### DIFF
--- a/test/tst_xdgdesktopfile.cpp
+++ b/test/tst_xdgdesktopfile.cpp
@@ -27,13 +27,13 @@ class Language
 {
 public:
     Language (const QString& lang)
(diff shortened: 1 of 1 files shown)


paulolieuthier:
What if `LC_MESSAGES` is not defined?

luis-pereira:
@paulolieuthier We set the `LC_MESSAGES` for the test. The order I was talking abut is [this one](https://github.com/lxde/libqtxdg/blob/master/qtxdg/xdgdesktopfile.cpp#L680).

luis-pereira:
Rebased. Closed by c39a9a7ad2503ac00f0c012e606bec897a1f30e5.

PR #106, opened 2016-09-29
Remove cpack

agaida:
We will not use it anymore


agaida:
@luis-pereira @palinek @paulolieuthier @jleclanche @pmattern - what do you think? - With .gitattributes we can use git archive without problems and don't have the export code in CMakeLists.txt lxqt-wide

jleclanche:
I doN't think it was ever really useful, tbh.

luis-pereira:
@agaida You know better than me. If you think it's better, go ahead.

paulolieuthier:
Pull the trigger.

palinek:
I'm also out of the discussion because lack of knowledge...

PR #109, opened 2016-10-03
Allow xdg-user-dirs in the realpath of $HOME.

jsm222:
On some systems /home is a symlink and $HOME points to the symlink.
This commit allows the xdg-user-dirs to start with the real/canonical path.

This did happen to be a problem in FreeBSD when changing xdg-user-dirs wiith lxqt-session-config  [lxqt-session-config](https://github.com/lxde/lxqt-session/blob/master/lxqt-config-session/userlocationspage.cpp#L175) returned false, because the path started with /usr and was not in $HOME..

### DIFF
--- a/qtxdg/xdgdirs.cpp
+++ b/qtxdg/xdgdirs.cpp
@@ -178,10 +178,12 @@ bool XdgDirs::setUserDir(XdgDirs::UserDirectory dir, const QString& value, bool 
     if (dir < XdgDirs::Desktop || dir > XdgDirs::Videos)
         return false;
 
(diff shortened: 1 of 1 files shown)


jsm222:
I've added `const` type qualifier to the `home` variable.
It seemed you forgot in your reivew.

luis-pereira:
@jsm222 No. That was me asking you to do it ;)

luis-pereira:
@jsm222 Thanks. Landed at https://github.com/lxde/libqtxdg/commit/65a412fdf1def3569c77c576055d0a35d7dc8fd9

PR #112, opened 2016-11-02
Refactors superbuild support

luis-pereira:
Use CMake policy NEW CMP0024 behavior.
Reference: https://cmake.org/cmake/help/v3.0/policy/CMP0024.html


agaida:
Right :) GTM

PR #115, opened 2017-02-05
Really return fallback icon

tsujan:
When an icon is missing, `XdgIcon::fromTheme()` returns another icon whose name is the starting part of the missing icon name, instead of showing fallback. This patch fixes that and will be followed by another patch for lxqt-panel.

### DIFF
--- a/qtxdg/xdgicon.cpp
+++ b/qtxdg/xdgicon.cpp
@@ -104,7 +104,7 @@ QIcon XdgIcon::fromTheme(const QString& iconName, const QIcon& fallback)
 
     // Note the qapp check is to allow lazy loading of static icons
     // Supporting fallbacks will not work for this case.
-    if (qApp && !isAbsolute && (icon.availableSizes().isEmpty() || icon.name() != iconName))
+    if (qApp && !isAbsolute && icon.availableSizes().isEmpty())
     {
         return fallback;
     }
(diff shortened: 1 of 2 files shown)


palinek:
This violates the [XDG specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html):
```
The dash “-” character is used to separate levels of specificity in icon names, for all contexts other than MimeTypes. For instance, we use “input-mouse” as the generic item for all mouse devices, and we use “input-mouse-usb” for a USB mouse device. However, if the more specific item does not exist in the current theme, and does exist in a parent theme, the generic icon from the current theme is preferred, in order to keep consistent style.
```

tsujan:
I changed the whole PR. And yes, I know about the "dash rule". But THAT WAS THE CAUSE OF ALL PROBLEMS in libqtxdg. I removed it **ffor now** -- everything is OK.

This PR is NOT finished yet. I'm trying to restore the "dash rule" in a meaningful way. The current implementation isn't acceptable. This is the correct way:

When searching for "preferences-system-session-services":

(1) First see if it exists in an `inherited` theme;
(2) And only if it isn't found in an inherited theme, use "preferences-system-session", or "preferences-system",....
,,,,

tsujan:
@palinek OK, it says: "if the more specific item does not exist in the current theme, *and does exist in a parent theme*, the generic icon from the current theme is preferred..." This isn't acceptable IMO, because when a more specific icon is found in the inherited theme, why not using it? XDG specification answers, "in order to keep consistent style," but in that case, the missing icons will be never found -- I say that as someone who has made some icon sets.

Anyway, it seems that I have to agree with the rule because I'm alone in this. Closing the PR.

PR #116, opened 2017-02-05
Correct hierarchy of fallbacks

tsujan:
This is my idea of a *practically* correct hierarchy of fallbacks, in contrast to what XDG specification tells us.

The icon is searched in the theme *and all of its parents*. Then, if it is not found, the search is done for the first "dash fallback" in the theme and all of its parents. And so on for the second, third,... "dash fallbacks".

Anyone has right to hate this but I use it to have acceptable fallback icons under LXQt, in the same way I have them under KDE.

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -295,9 +295,32 @@ XdgIconTheme::XdgIconTheme(const QString &themeName)
 #endif //QT_NO_SETTINGS
 }
 
+/* WARNING:
+ *
+ * https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html
+ *
+ * <cite>
+ * The dash “-” character is used to separate levels of specificity in icon
+ * names, for all contexts other than MimeTypes. For instance, we use
+ * “input-mouse” as the generic item for all mouse devices, and we use
+ * “input-mouse-usb” for a USB mouse device. However, if the more specific
+ * item does not exist in the current theme, and does exist in a parent
+ * theme, the generic icon from the current theme is preferred, in order
+ * to keep consistent style.
+ * </cite>
+ *
+ * But we believe, that using the more specific icon (even from parents)
+ * is better for user experience. So we are violating the standard
+ * intentionally.
+ *
+ * Ref.
+ * https://github.com/lxde/lxqt/issues/1252
+ * https://github.com/lxde/libqtxdg/pull/116
+ */
 QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
                                  const QString &iconName,
-                                 QStringList &visited) const
+                                 QStringList &visited,
+                                 bool dashFallback) const
 {
     QThemeIconInfo info;
     Q_ASSERT(!themeName.isEmpty());
@@ -385,17 +408,10 @@ QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
             }
         }
 
-        if (!info.entries.isEmpty()) {
+        if (!info.entries.isEmpty())
             info.iconName = iconNameFallback.toString();
-            break;
-        }
 
-        // If it's possible - find next fallback for the icon
-        const int indexOfDash = iconNameFallback.lastIndexOf(QLatin1Char('-'));
-        if (indexOfDash == -1)
-            break;
-
-        iconNameFallback.truncate(indexOfDash);
+        break;
     }
 
     if (info.entries.isEmpty()) {
@@ -480,6 +496,16 @@ QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
     }
 #endif
 
+    if (dashFallback && info.entries.isEmpty()) {
+        // If it's possible - find next fallback for the icon
+        const int indexOfDash = iconNameFallback.lastIndexOf(QLatin1Char('-'));
+        if (indexOfDash != -1) {
+            iconNameFallback.truncate(indexOfDash);
+            QStringList _visited;
+            info = findIconHelper(themeName, iconNameFallback.toString(), _visited, true);
+        }
+    }
+
     return info;
 }
 
@@ -488,7 +514,7 @@ QThemeIconInfo XdgIconLoader::loadIcon(const QString &name) const
     const QString theme_name = QIconLoader::instance()->themeName();
     if (!theme_name.isEmpty()) {
         QStringList visited;
-        return findIconHelper(theme_name, name, visited);
+        return findIconHelper(theme_name, name, visited, true);
     }
 
     return QThemeIconInfo();
--- a/xdgiconloader/xdgiconloader_p.h
+++ b/xdgiconloader/xdgiconloader_p.h
@@ -134,7 +134,8 @@ public:
 private:
     QThemeIconInfo findIconHelper(const QString &themeName,
                                   const QString &iconName,
-                                  QStringList &visited) const;
+                                  QStringList &visited,
+                                  bool dashFallback = false) const;
     mutable QStringList m_iconDirs;
     mutable QHash <QString, XdgIconTheme> themeList;
 };


luis-pereira:
@tsujan It seems OK to me. This change can have unforeseen consequences. I will test it for a couple of days.

agaida:
@tsujan @luis-pereira  i will use this too and go a step further - if ok for me i will add this to the fixes for Siduction after a few days to get more feedback eventually. We don't need to hurry right now but we should finally fix it - if tested enough and considered ok.

tsujan:
@luis-pereira @agaida Thanks!

palinek:
I personally have no opinion on this => both approaches are logical.

But we should mention, that we are violating the standard intentionally. At least in the source code like e.g. [here](https://github.com/lxde/libqtxdg/blob/master/qtxdg/xdgdesktopfile.cpp#L372)

tsujan:
To be clear as far as poossible, I explain this with math symbols and an example. Supposing that X and Y are icon sets and X is active, this is what XDG specification gives us:
````
(X ⊂ Y) ∧ (gnome-disks ∉ X) ∧ (gnome ∈ X) ∧ (gnome-disks ∈ Y) ⇒ (gnome ∈ X)
````
And this is what the PR does:
````
(X ⊂ Y) ∧ (gnome-disks ∉ X) ∧ (gnome ∈ X) ∧ (gnome-disks ∈ Y) ⇒ (gnome-disks ∈ Y)
````
Now, `gnome ∈ X` may have the general look of other icons in X but is meaningless in this case. On the other hand, `gnome-disks ∈ Y` may not have the general look of other X icons but shows what this app really is.

The responsibility of `gnome-disks ∉ X` is on the shoulders of the creator of X;  we shouldn't compensate for the incompleteness of an icon set with a meaningless icon -- that wouldn't be logical -- but try to provide the user with a meaningful one.

**Comment by agaida** [comment]
@tsujan - sorry for being so silent the last days - build in all the PRs and have to sort some things out, still not finished

**Comment by tsujan** [comment]
@agaida Take you time! I'm busy with the latest changes in libfm-qt (and also Dolphin's code, ...).

**Comment by luis-pereira** [comment]
@tsujan I'm in the process of evaluating this. First of all, sorry for the delay.
Can you point an example where KDE does the same thing ?

**Comment by tsujan** [comment]
@luis-pereira 
Please read https://github.com/lxde/libqtxdg/pull/116#issuecomment-278041342. It includes a real example.

**Comment by luis-pereira** [comment]
@tsujan Thanks. What `X` and `Y` illustrates KDE behaving that way ?

p.s. I'm making some "practical" tests with `kiconfinder` and `qtxdg-iconfinder`.

**Comment by tsujan** [comment]
@luis-pereira 
KDE behaves in the way of this PR.

You could:
(1) Install gnome-disks;
(2) Add an icon named `gnome` to the icon themes you use;
(3) Remove `gnome-disks` from that theme;
(4) Log out and in.

**Comment by luis-pereira** [comment]
@tsujan Under KDE with the Human icon theme what's the output of:
`kiconfinder system-suspend` ?
and under LXQt also with the Human theme and this PR what's the output of:
`qtxdg-iconfinder system-suspend` ?

p.s. You have to rebase this PR to get the `qtxdg-iconfinder` tool.

**Comment by tsujan** [comment]
@luis-pereira 

`kiconfinder` correctly finds "/usr/share/icons/Human/48x48/devices/system.png" because Human inherits Tangerine and gnome, I don't have Tangerine, and neither Human nor gnome has "system-suspend".

I work under KDE for a few weeks (to study and patch Dolphin's code,...) and can't go to LXQt for now; so I don't have `qtxdg-iconfinder` yet but I'd tested this PR in several different ways *before* making it and it had worked fine. It's logic is clear too. Do you have a real example, where it doesn't work? I don't know about  `qtxdg-iconfinde` yet but remember that icon cache sometimes played tricks, so that, to see the real icons, I had to log out and in.

**Comment by tsujan** [comment]
I forgot to add that, as in the case of my other PR's that aren't related to pcmanfm-qt, libfm-qt, etc., whatever you or @palinek do with this PR (editing, rebasing adding comments,...) will be OK to me. But please note that LXQt's current way of handling icons is incorrect -- in fact, the so-called "standards" are incorrect in this case --  and this PR is a step to make it right.

**Comment by luis-pereira** [comment]
@tsujan 
I delete all cache files before experimenting.
Take a look at `qtxdg-iconfinder` with this PR applied. It does not behave like the KDE engine.
```
qtxdg-iconfinder system-suspend
system-suspend:system-suspend
        /usr/share/icons/oxygen/base/128x128/actions/system-suspend.png
        /usr/share/icons/oxygen/base/22x22/actions/system-suspend.png
        /usr/share/icons/oxygen/base/32x32/actions/system-suspend.png
        /usr/share/icons/oxygen/base/64x64/actions/system-suspend.png
        /usr/share/icons/oxygen/base/16x16/actions/system-suspend.png
        /usr/share/icons/oxygen/base/48x48/actions/system-suspend.png
```
without this PR the output is following:
```
qtxdg-iconfinder system-suspend
system-suspend:system
        /usr/share/icons/Human/48x48/devices/system.png
        /usr/share/icons/Human/24x24/devices/system.png
        /usr/share/icons/Human/16x16/devices/system.png
        /usr/share/icons/Human/scalable/devices/system.svg
```
which matches KDE icon engine behavior.

**Comment by tsujan** [comment]
@luis-pereira 
That seems impossible to me. A bug in qtxdg-iconfinder? Which icon is really shown on the menu (after log-out/log-in)?

**Comment by tsujan** [comment]
@luis-pereira 
You made me log into LXQt and see for myself ;) KDE does it wrong; this PR does it right! Why? Because this is in Tangerine's `index.theme`:
````
Inherits=gnome,oxygen
````
Human inherits Tangerine's.

Why KDE doesn't go deeper, I don't know but this PR takes *all parents* into account correctly :)

**Comment by tsujan** [comment]
In simple word, finding and showing that oxygen icon was the whole point of the PR: showing a meaningful icon as far as possible. Your test proved that.

I think KDE sets a limit to the number of parents that should be searched (only the first `index.theme`?).

**Comment by luis-pereira** [comment]
@tsujan
[`qtxdg-iconfinder`](https://github.com/lxde/libqtxdg/blob/master/util/qtxdg-iconfinder.cpp) is really simple. It's results are trustworthy.
I'm trying to have an informed opinion. Before breaking deliberately the standard I want to be in possession of "all" facts.   
You stated that KDE also breaks the standard in this regard. While I think that what others do or not do is not that relevant, it may point some issues.

Iterating all the parents of a theme, can bring a performance penalty. Specially when theme creators abuse the `Inherits` entry.

Can you edit the Tangerine theme to:
`Inherits=oxygen` and run again `kiconfinder system-suspend` ?

**Comment by luis-pereira** [comment]
Rebased

**Comment by tsujan** [comment]
> qtxdg-iconfinder is really simple. It's results are trustworthy.

I know. As I wrote above, I wrongly trusted KDE. Everything is OK.

> Iterating all the parents of a theme, can bring a performance penalty. Specially when theme creators abuse the Inherits entry.

I vaguely remember that the method in question was protected against infinite loops. Sorry, I'm very busy with another code; would you please check that? Apart from it, there can't be any penalty; we don't have 100 parents ;)

> Can you edit the Tangerine theme to:....

Sorry, I can't, for the above reason.

IMHO, you could acquire the certainty you seek by reading the code. All my tests and also your test -- that was supposed to be a counter-example -- have confirmed that the code works fine.

**Comment by tsujan** [comment]
OK. I came back to LXQt from KDE. Everything feels much lighter now :D

I took a look at the code again and found nothing wrong. Iterating over all the parents of a theme is what should be done; otherwise the keyword `Inherits` would have no meaning. Those who make icon sets use `Inherits` for a reason. Moreover, iterating over all parents isn't something this PR has introduced. It was already there; the difference is just in the method of iteration.

I think we discussed this too much. I'll listen to any rational criticism or reasoning but I'm afraid that too much of cautiousness is rather a personal matter.

**Comment by luis-pereira** [comment]
@tsujan your code does what it promises. There's no doubt about that.
I'm in a trip. Will come back later today.

**Comment by agaida** [comment]
@tsujan - i'm nearly fine with the changes - it might be that i'm overly picky or understand something wrong so please bare with me if i'm missing the point - but i have a visual problem with the impact of the changes :) 
screen10
^^ For the icons this is the outcome i would expect

screen11
^^ and it would be nice to have that matching too - in this special case it would point to another problem  - the leave menu should use the system theme which is a modified frost in my case.

**Comment by tsujan** [comment]
@agaida I'll come to my computer soon but until then, could you please explain what's the problem? I didn't get it.

**Comment by tsujan** [comment]
@agaida, I'm back.
Supposing that you use this PR, with the first screenshot, you've chosen Breeze. Let's take the log-out icon as an example. Its name is "system-log-out". In my system (latest Manjaro Testing) Breeze has this (ugly) icon for "system-log-out":

system-log-out

And it's displayed correctly.

In the second screenshot, you've chosen Breeze Dark and it has this (uglier) icon  for "system-log-out":

system-log-out1

So, your second screenshot can't be correct ;) In my system, Andromeda has that icon.

> the leave menu should use the system theme which is a modified frost in my case

But you've chosen Breeze Dark in your second screenshot. Maybe that's an example of icon cache trick!?

**Comment by agaida** [comment]
@tsujan - you are right, i would expect the fugly icon - but i get the colorful icons - any ideas where i can search? Hmm, i guess i should work a little bit on a pure LXQt liveiso with latest snapshot packages. :)

**Comment by tsujan** [comment]
@agaida, even without this PR you shouldn't get those colored icons when you switch to Breeze Dark because it doesn't have them -- at least, not here, with breeze-icons 5.32.0. I know they belong to Andromeda but another theme (inheriting Andromeda) may have them too.

**Comment by tsujan** [comment]
BTW, Breeze Dark is for dark widget styles. Its icons won't be easily visible on light backgrounds under LXQt. KDE has an interesting feature LXQt lacks: when SVG icons are made appropriately -- as Breeze sets are -- they become white on (some) dark backgrounds and black on light ones. I know the basics of the code but not all of it.

**Comment by agaida** [comment]
First - the PR is ok, second - i'm getting old and grumpy, third - the expected icons are not in debians dark theme - for what reason ever. And i have no clue, which icons are displayed nor why, had to search it in Numix or in hicolor i guess.

**Comment by tsujan** [comment]
> i'm getting old and grumpy

Join the club ;)

> it would be cool if we could have some automatic

You mean https://github.com/lxde/lxqt/issues/1181 ? It's tricky but quite possible.

> or that it is possible to set icon themes on a application base

It isn't impossible.

**Comment by agaida** [comment]
and here is the reason for the nice coloured icons:
screen12
the three files that debian deliver:> icons/breeze-dark % find . | grep system-log-out                                
> ./apps/16/system-log-out.svg
> ./apps/22/system-log-out.svg
> ./apps/32/system-log-out.svg

**Comment by tsujan** [comment]
> and here is the reason for the nice coloured icons:

Oh, that's 5.28.0 (I downloaded it from https://packages.debian.org/sid/breeze-icon-theme right now). The colored icon is replaced by a colorless one in 5.32.0. So, you second screenshot showed a correct icon.

**Comment by agaida** [comment]
sometimes i really love linux development - it's so predictable ...

**Comment by agaida** [comment]
ok, works fine with the 5.32 icons - and thats a bad thing ;) - we really need to adjust our themes systemwide or provide a mechanism to set different icon themes or something like lxde/lxqt/#1181 (or both).
screen14
screen15
screen16

**Comment by tsujan** [comment]
> we really need to adjust our themes systemwide or provide a mechanism to set different icon themes ...

That's not needed. Breeze is tuned to the feature mentioned at https://github.com/lxde/lxqt/issues/1181. That feature needes 3 things: (1) LXQt code; (2) An icon set like Breeze (whose SVG icons have a special preperty; and (3) a style engine that supports the feature (currently only Kvantum and Breeze do). We need to do (1).

**Comment by agaida** [comment]
Indeed: From this point of view it isn't needed - but it would be nice to have.

**Comment by tsujan** [comment]
This is a screenshot of Dolphin with the Breeze icon set and the Kvantum Style engine (with its theme KvSimplicity) under KDE:

kvantum

See how breeze icons change color on selected menu-items: they become white because the background is dark blue. The same thing doesn't happen under LXQt because LXQt still lacks this nice feature.

**Comment by agaida** [comment]
go for it :)

**Comment by tsujan** [comment]
Sure; just a little more free time. I think I'll have it soon.

**Comment by agaida** [comment]
Don' hurry - if we solve all problems today, what should we do tomorrow? That would be boring ...

**Comment by tsujan** [comment]
:)

To prove that it isn't specific to KDE apps, this is pcmanfm-qt with Breeze icons and another Kvantum theme but under KDE:

pcmanfm

This time, icons are white on the dark toolbar and selected menu-item but white on the light menu. The default Breeze set is used here.

**Comment by agaida** [comment]
@tsujan @palinek @luis-pereira  - any objections to merge this?

**Comment by tsujan** [comment]
As the PR maker, I can't have any objection, of course ;)

**Comment by palinek** [comment]
As I've said, I have no problem with either of the prioritization logic. But we definitely should mention (in doc?, source code comment? ..), that we are violating the standard for a good reason.

**Comment by tsujan** [comment]
> But we definitely should mention (in doc?, source code comment? ..), that we are violating the standard for a good reason.

I can't find a good place in the source for such a comment.

**Comment by agaida** [comment]
If anything else fails - README seems a good place

**Comment by tsujan** [comment]
An explanation could contain a statement like this (but, perhaps, not exactly this):

"... The XXX standard gives priority to the current icon set, sometimes at the cost of a meaningless icon, while YYY searches for a meaningful icon first... "

**Comment by agaida** [comment]
We should put the explanation in a separated Pull Request - such explanations have a great potentioal for bike shedding.

**Comment by palinek** [comment]
> > But we definitely should mention (in doc?, source code comment? ..), that we are violating the standard for a good reason.

> I can't find a good place in the source for such a comment.

Note in (or above) the `findIconHelper()`?:
```c++
  /* WARNING:
   *
   * https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html
   *
   * <cite>
   * The dash “-” character is used to separate levels of specificity in icon
   * names, for all contexts other than MimeTypes. For instance, we use
   * “input-mouse” as the generic item for all mouse devices, and we use
   * “input-mouse-usb” for a USB mouse device. However, if the more specific
   * item does not exist in the current theme, and does exist in a parent
   * theme, the generic icon from the current theme is preferred, in order
   * to keep consistent style.
   * </cite>
   *
   * But we believe, that using the more specific icon (even from parents)
   * is better for user experience. So we are violating the standard
   * intentionally.
   *
   * Ref.
   * https://github.com/lxde/lxqt/issues/1252
   * https://github.com/lxde/libqtxdg/pull/116
   */
```

agaida:
:+1:

tsujan:
@palinek 
Yes, that would be good. I'll add it tomorrow or, if you have time now, could you add it?

palinek:
done

tsujan:
@palinek 2 min ago I added it and wanted to push but saw your comment!!

Should it be merged now?

luis-pereira:
@agaida I do have issues with it that are still not clear to me.
There are some statements in this PR that are unproven **to me**.
I did some quick load time measurements and I found load times roughly 3x times higher.
Will try to make systematize it ASAP.

If in the meantime this goes in... that's life.

agaida:
@luis-pereira - we are not in hurry right now - and qtxdg is a central part of LXQt - so please take your time. We will wait with the merge.

luis-pereira:
@tsujan A rebase would be nice.

tsujan:
@luis-pereira 
But it has no conflict with anything yet.

luis-pereira:
@tsujan I know, but, github is (wrongly) detecting an merge conflict when I try to create a new PR onto this branch.

tsujan:
OK. It's after midnight here and I haven't slept for 48 hours. Will do it tomorrow.

luis-pereira:
Sure. I'm still jet-legging.... and will also go to sleep.

tsujan:
I rebased it a few minutes ago because I might wake up very late tomorrow ;)

luis-pereira:
@tsujan Thanks.

agaida:
@tsujan @luis-pereira any news?

tsujan:
> @tsujan @luis-pereira any news?

@luis-pereira has his doubts. From our discussions above, I concluded that I'm not able to convince him that this is all right; so I won't try to.

luis-pereira:
@agaida @tsujan Although I think visual coherence is important and this PR breaks it, I'm Ok with it.
My only concern is about performance. Any comments on https://github.com/lxde/libqtxdg/issues/128 ?

tsujan:
> My only concern is about performance

OK, I really can't find anything to add besides what follows (not understanding what's so problematic with this simple change):

The performance obviously depends on the icon set. Supposing that `system-lock-screen` is needed:

(1) When a set and its parent don't have `system-lock-screen` but only `system-lock`, the "standard" method finds `system-lock` faster because it doesn't even try to search the parent for `system-lock-screen`.

(2) When a set only has `system` while its parent has `system-lock-screen`, `system-lock-screen` is found faster by this method than the meaningless icon `system` is found by the "standard" one.

agaida:
i guess i have an idea - i have an ancient Acer Pentium i386 Laptop somewhere in the basement - my last machine with real spinning drives. Will fire this machine up with latest sid and try to measure the differences. Sorry, even my machines buit in 2009 have SSDs.

tsujan:
> Sorry, even my machines built in 2009 have SSDs.

Ha ha... my old desktop computer is older than your old laptop -- I got it in 2008 ;)

luis-pereira:
@tsujan I've no problem with this.
My only point is to make sure that the performance penalty is acceptable on weak machines. The kind of machines that, for example, can't run KDE.

tsujan:
> My only point is to make sure that the performance penalty is acceptable

My point [above](https://github.com/lxde/libqtxdg/pull/116#issuecomment-296380447) is that there's no *absolute* performance penalty.

agaida:
OK - i guess we should merge it right now - once it is merged and built i will be able to set up two really old machines without to much effort needed (manual builds with a set of patches for i386 are really pita )

PR #117, opened 2017-02-16
Add iconfinder

luis-pereira:
It helps testing and evaluating https://github.com/lxde/libqtxdg/pull/116.

### DIFF
--- a/util/qtxdg-iconfinder.cpp
+++ b/util/qtxdg-iconfinder.cpp
@@ -1,0 +1,60 @@
+/*
+ * libqtxdg - An Qt implementation of freedesktop.org xdg specs
+ * Copyright (C) 2017  Luís Pereira <luis.artur.pereira@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA  02110-1301  USA
+ */
+
+#include <QGuiApplication> // XdgIconLoader needs a QGuiApplication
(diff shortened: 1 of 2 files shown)


util/qtxdg-iconfinder.cpp
```diff
@@ -0,0 +1,63 @@
...
+    parser.addVersionOption();
+    parser.addHelpOption();
+    parser.process(app);
+
+    if (parser.positionalArguments().isEmpty())
+        parser.showHelp(EXIT_FAILURE);
+
+    const QStringList icons = parser.positionalArguments();
+    for (const QString& iconName : icons) {
+        const auto icon = XdgIconLoader::instance()->loadIcon(iconName).iconName;
```
palinek: I would call the `XdgIconLoader::loadIcon()` only once...

util/qtxdg-iconfinder.cpp
```diff
@@ -0,0 +1,63 @@
...
+        const auto entries = XdgIconLoader::instance()->loadIcon(iconName).entries;
+
+        std::cout << qPrintable(iconName) <<
+            qPrintable(QString::fromLatin1(":")) << qPrintable(icon) << "\n";
+
+        if (!icon.isEmpty()) {
+            if (!entries.empty()) {
```
palinek: Is the `empty()` check necessary?

util/qtxdg-iconfinder.cpp
```diff
@@ -0,0 +1,63 @@
...
+                const auto end = entries.constEnd();
+                for (auto i = entries.constBegin(); i != end; ++i) {
```
palinek: Just curious, why are you using the for range loop on one place and manual iteration on other?

util/qtxdg-iconfinder.cpp
luis-pereira: Me too. I added the entries stuff later, and .... bed prevent me from compacting it.

util/qtxdg-iconfinder.cpp
luis-pereira: Reasoning in the former comment.

util/qtxdg-iconfinder.cpp
luis-pereira: I'm trying to adopt the range for and other C++11 stuff... but years of using the manual iterators "sometimes" kick in, without I even notice it.

luis-pereira:
@palinek Updated.

util/qtxdg-iconfinder.cpp
palinek: I meant the `if (!entries.empty())` ...because if the `entries` are empty, the for loop will do nothing.

luis-pereira:
@palinek Thanks. Jet-lag is killing me.

PR #118, opened 2017-03-13
xdgiconloader: Implement QIconEnginePlugin interface

palinek:
There was no way to serialize-deserialize QIcon to-from QDataStream,
because the `QDataStream &operator>>(QDataStream &s, QIcon &icon)` does
use the QIconEnginePlugin interface to load specific (user-defined)
QIconEngines.

This implements the "factory" to create XdgIconLoaderEngine objects
based on the serialized type/key "XdgIconLoaderEngine" (provided for
serialization by `XdgIconLoaderEngine::key()`).

The problem was revealed by investigation of martinrotter/rssguard#95

@luis-pereira please, review this... as it seem to be a major drawback of xdgiconloader for now

@agaida is this OK from the "packagers point of view"?


pmattern:
ac7ff4f is fixing martinrotter/rssguard#95 here. Cannot judge the technical details.

luis-pereira:
@palinek 
1. Plugins should use the `MODULE` type. Documentation [here](https://cmake.org/cmake/help/v3.0/command/add_library.html) Consider this:
``` diff
diff --git a/xdgiconloader/plugin/CMakeLists.txt b/xdgiconloader/plugin/CMakeLists.txt
index 70bae7b..9682d6d 100644
--- a/xdgiconloader/plugin/CMakeLists.txt
+++ b/xdgiconloader/plugin/CMakeLists.txt
@@ -2,7 +2,7 @@ set(xdgiconengineplugin_CPP_FILES
     xdgiconengineplugin.cpp
 )
 
-add_library(${QTXDGX_ICONENGINEPLUGIN_LIBRARY_NAME} SHARED
+add_library(${QTXDGX_ICONENGINEPLUGIN_LIBRARY_NAME} MODULE
     ${xdgiconengineplugin_CPP_FILES}
 )
```

2. What about making `QTXDGX_ICONENGINEPLUGIN_INSTALL_PATH` an option ?

luis-pereira:
Sorry, closed by mistake, obviously.

palinek:
> 1. Plugins should use the `MODULE` type. 

right, done

> 2. What about making `QTXDGX_ICONENGINEPLUGIN_INSTALL_PATH` an option ?

But how? [option](https://cmake.org/cmake/help/v3.1/command/option.html) can just be an ON/OFF switch.

luis-pereira:
> But how? option can just be an ON/OFF switch.

Of course. I meant [marking it as advanced](https://cmake.org/cmake/help/v3.0/command/mark_as_advanced.html).

palinek:
> Of course. I meant marking it as advanced.

I'm not familiar with it and don't see any benefit of using it right now... so please, make needed changes here...

luis-pereira:
@palinek Done. It's probably better to squash these tiny tiny commits.

palinek:
OK. We're finished now. Are we going to wait for @agaida (as he probably has the best packager perspective) to approve this?

luis-pereira:
@palinek You can pack it all in one commit, if you want.
I do agree with waiting for @agaida.

agaida:
will have a look today - sorry, was very busy in the last weeks

palinek:
> will have a look today

Great. Also, please, consider release... as IMO this deserves it.

agaida:
From  packaging POV no problem, applies fine, work fine for me™ with no drawbacks - so i would give it a go. Regarding the new release: I'm fine with, but i would like to have #116 in for that release too.

PR #119, opened 2017-03-28
Support symbolic SVG icons

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1181.

That means changing their colors so that they always have a high contrast with their backgrounds. Symbolic icons should have SVG stylesheets (as the Breeze icons do) and the style engine should support them (as the Breeze and Kvantum styles do).

The code is stolen from KDE's `kiconloader.cpp` (see the comments).

Currently Panel and those LXQt apps that use stylesheets will show the wrong icon color if their stylesheets have a high contrast with the active Qt style. This issue should be fixed in their codes before this PR is merged.

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -719,13 +719,10 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
 
 QPixmap ScalableEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state)
 {
-    if (svgIcon.isNull())
-        svgIcon = QIcon(filename);
-
-    QImage img;
     // The following lines are adapted from KDE's "kiconloader.cpp" ->
     // KIconLoaderPrivate::processSvg() and KIconLoaderPrivate::createIconImage().
     // They read the SVG color scheme of SVG icons and give images based on the icon mode.
+    QImage img;
     QScopedPointer<QImageReader> reader;
     QBuffer buffer;
     if (filename.endsWith(QLatin1String("svg"))) {
@@ -771,17 +768,17 @@ QPixmap ScalableEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State 
     }
 
     QPixmap px;
-    QIcon icn;
     if (!img.isNull() && px.convertFromImage(img)) {
         // Do not return the pixmap now but get the icon from it
         // for QIcon::pixmap() to handle states and modes,
         // especially the disabled mode.
-        icn = QIcon(px);
+        svgIcon = QIcon(px);
     }
-    else
-      icn = svgIcon;
 
-    return icn.pixmap(size, mode, state);
+    if (svgIcon.isNull()) // not really needed with SVG icons
+        svgIcon = QIcon(filename);
+
+    return svgIcon.pixmap(size, mode, state);
 }
 
 QPixmap XdgIconLoaderEngine::pixmap(const QSize &size, QIcon::Mode mode,
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -45,6 +45,9 @@
 #include <QtCore/QDir>
 #include <QtCore/QSettings>
 #include <QtGui/QPainter>
+#include <QImageReader>
+#include <QXmlStreamReader>
+#include <QBuffer>
 
 #ifdef Q_DEAD_CODE_FROM_QT4_MAC
 #include <private/qt_cocoa_helpers_mac_p.h>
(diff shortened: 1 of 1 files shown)


palinek:
1. the SVG processing should be done only in case when the theme states it (`FollowsColorScheme=true`)

palinek:
@tsujan I can have a look in those 2 mentioned points... if you wish

tsujan:
> I can have a look in those 2 mentioned points... if you wish

That would be great. Thanks!

> the SVG processing should be done only in case when the theme states it 

I disagree and think it's a redundant KDE check. KDE has the concept of "color scheme"; we don't (I remember that I explained that somewhere a year ago).

Yes, we stole the code from them but that doesn't mean we should follow them in each step. On the one hand, this isn't a CPU-intensive process for user to have an option about it. On the other hand, most KDE users don't know about "FollowsColorScheme" and If an SVG icon set doesn't have SVG stylesheets, it won't follow the app palette (the so-called KDE "color scheme") automatically.

tsujan:
A long story made short, I mean that "FollowsColorScheme" is a KDE concept.

palinek:
> If an SVG icon set doesn't have SVG stylesheets, it won't follow the app palette (the so-called KDE "color scheme") automatically.

But the whole XML processing is needless. Therefore IMO it is good to follow the "FollowsColorScheme"... even if it is an KDE extension.

tsujan:
`xmlReader.qualifiedName() == QLatin1String("style")` and `xmlReader.attributes().value(QLatin1String("id")) == QLatin1String("current-color-scheme")` would be false without SVG stylesheet and I saw no difference with my own SVG icons, which don't have stylesheets. The code is really fast.

tsujan:
As a matter of fact, the process never reaches these lines, not even with svg icons without stylesheet:
````
    }
    else
        icn = svgIcon;
````
So, there's no redundant calculation.

**Comment by palinek** [comment]
> The code is really fast.

It may be in global scope, but
1. It does depend on the structure of the XML
2. This (needless) processing (XML DOM parsing) gives our engine performance penalty. (I'm sure you agree, that checking one bool variable must be faster than parsing whole XML and checking some DOM nodes)

**Comment by tsujan** [comment]
Please read my last comment above yours. We commented simultaneously ;)

**Comment by tsujan** [comment]
In other words, this is needed without stylesheet:
````
                else if (xmlReader.tokenType() != QXmlStreamReader::Invalid)
                    writer.writeCurrentToken(xmlReader);
````
No redundant processing!

**Comment by tsujan** [comment]
>  checking one bool variable must be faster than parsing whole XML and checking some DOM nodes

Even without this patch, the svg was processed as an xml file. The detail just wasn't in our code.

**Comment by palinek** [comment]
>> checking one bool variable must be faster than parsing whole XML and checking some DOM nodes

> Even without this patch, the svg was processed as an xml file. The detail just wasn't in our code.

Of course, but now there is one XML parsing more.

**Comment by tsujan** [comment]
Which parsing do you mean?

**Comment by palinek** [comment]
> In other words, this is needed without stylesheet:
```
                else if (xmlReader.tokenType() != QXmlStreamReader::Invalid)
                    writer.writeCurrentToken(xmlReader);
```
No redundant processing!

And this is completely wrong assumption -> the XML parser must have been read/parsed the file

tsujan:
@palinek 
Remove `icn = svgIcon;` and you'll see svg icons without stylesheet will be shown correctly. Then also remove
````
                else if (xmlReader.tokenType() != QXmlStreamReader::Invalid)
                    writer.writeCurrentToken(xmlReader);
````
and you'll see no icon.

palinek:
> Remove icn = svgIcon; and you'll see svg icons without stylesheet will be shown correctly. Then also remove

I'm not saying, that it's not working. I'm just saying, that the processing is needles for icons from themes, that do not "FollowsColorScheme"

tsujan:
> the XML parser must have been read/parsed the file

Where?

palinek:
... just had a look into the doc. The `QXmlStreamReader` is an SAX parser (not DOM)... so better but the processing is still there.

tsujan:
````
QXmlStreamReader is a faster and more convenient replacement for Qt's own SAX parser.... In some cases it might also be a faster and more convenient alternative for use in applications that would otherwise use a DOM tree ....
````

tsujan:
There's one thing. I think this line
````
     if (svgIcon.isNull())
          svgIcon = QIcon(filename);
````
might be moved and become
````
     if (svgIcon.isNull())
          svgIcon = icn;
````

tsujan:
More specifically, these lines before `return`:
````
    if (svgIcon.isNull()) {
        if (!icn.isNull())
            svgIcon = icn;
        else {
            svgIcon = QIcon(filename);
            icn = svgIcon;
        }
    }
````

tsujan:
Or:
````
    QPixmap px;
    if (!img.isNull() && px.convertFromImage(img)) {
        // Do not return the pixmap now but get the icon from it
        // for QIcon::pixmap() to handle states and modes,
        // especially the disabled mode.
        svgIcon = QIcon(px);
    }

    if (svgIcon.isNull())
        svgIcon = QIcon(filename);

    return svgIcon.pixmap(size, mode, state);
````
*This* means that there's a redundant reading for icons without stylesheet because those SVG files are read again when `svgIcon` isn't null, while there's no need to read them again because their color won't change. In this sense, the code should be changed. `FollowsColorScheme` would be the last resort.

tsujan:
OK! You're right. My hypotheses were wrong. `svgIcon` is parsed before. I'll try to add `FollowsColorScheme`.

palinek:
> OK! You're right. My hypotheses were wrong. svgIcon is parsed before. I'll try to add FollowsColorScheme.

I'm allready on it... but found some problem outside this PR.

tsujan:
> I'm allready on it... but found some problem outside this PR.

Great! Let's work on it independently and see what comes out.

tsujan:
@palinek 
I think the code should be placed inside `XdgIconLoaderEngine::pixmap()`. Then, `FollowsColorScheme` could be easily respected. I'll add a commit after testing.

tsujan:
All right. I succeed to consider  `FollowsColorScheme` by placing the code inside `XdgIconLoaderEngine::pixmap()` and using `XdgIconLoader::instance()->theme()`.

However, there's a more important matter. If `FollowsColorScheme` is `true`, the code shouldn't create `svgIcon` and then correct it. It seems that the structure of  `xdgiconloader.cpp` should be changed, although I prefer to be wrong in this.

My brain's temperature has got high. Will think about it later, hoping that @palinek might find a simple solution.

tsujan:
Superseded by https://github.com/lxde/libqtxdg/pull/121

PR #120, opened 2017-03-28
Support symbolic SVG icons V2

tsujan:
This is https://github.com/lxde/libqtxdg/pull/119 plus `FollowsColorScheme`.

PR #121, opened 2017-03-28
Icon follow color

palinek:
@tsujan please, consider this one as follow-up of #119 (my implementation of FollowColorScheme and (big?) optimization of pixmap generation)

### DIFF
--- a/qtxdg/xdgicon.cpp
+++ b/qtxdg/xdgicon.cpp
@@ -145,6 +145,17 @@ QIcon XdgIcon::fromTheme(const QString &iconName,
     return fromTheme(icons);
 }
 
+bool XdgIcon::followColorScheme()
+{
+    return XdgIconLoader::instance()->followColorScheme();
+}
+
+
+void XdgIcon::setFollowColorScheme(bool enable)
+{
+    XdgIconLoader::instance()->setFollowColorScheme(enable);
+}
+
 
 QIcon XdgIcon::defaultApplicationIcon()
 {
--- a/qtxdg/xdgicon.h
+++ b/qtxdg/xdgicon.h
@@ -44,6 +44,16 @@ public:
                            const QString &fallbackIcon4 = QString());
     static QIcon fromTheme(const QStringList& iconNames, const QIcon& fallback = QIcon());
 
+    /*!
+     * Flag if the "FollowsColorScheme" hint (the KDE extension to XDG
+     * themes) should be honored. If enabled and the icon theme supports
+     * this, the icon engine "colorizes" icons based on the application's
+     * palette.
+     *
+     * Default is true (use this extension).
+     */
+    static bool followColorScheme();
+    static void setFollowColorScheme(bool enable);
     /* TODO: deprecate & remove all QIcon wrappers */
     static QString themeName() { return QIcon::themeName(); }
     static void setThemeName(const QString& themeName) { QIcon::setThemeName(themeName); }
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -76,6 +76,15 @@ static bool gSupportsSvg = false;
 static bool gSupportsSvg = true;
 #endif //QT_NO_LIBRARY
 
+void XdgIconLoader::setFollowColorScheme(bool enable)
+{
+    if (m_followColorScheme != enable)
+    {
+        QIconLoader::instance()->invalidateKey();
+        m_followColorScheme = enable;
+    }
+}
+
 XdgIconLoader *XdgIconLoader::instance()
 {
    QIconLoader::instance()->ensureInitialized();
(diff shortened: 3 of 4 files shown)


tsujan:
Interesting! At the first glance, this is very similar to one of my three solutions. But I'll read your code tomorrow, hoping that its readings are less that of https://github.com/lxde/libqtxdg/pull/120.

tsujan:
@palinek 
Before going to bed, I was so curious that I tested the branch `icon_follow_color`. I'm afraid it doesn't work. I haven't read the code yet (its general structure is like what I tested before though), so I have no clue about the cause.

Screenshot with this PR:



And this one with https://github.com/lxde/libqtxdg/pull/120:

palinek:
heh... maybe I've found the reason... the problem can be in the used stylesheet in the "generated" svg, because it is dependent on the QIcon::Mode...

tsujan:
Yes, it was `QIcon::Mode` that should be considered.

This is the best solution so far because both ordinary and selected icons are remembered and the readings are minimal. Well done!

You could just make the code tidier. For example, `m_followsColorTheme` and its related functions seem redundant in `XdgIconLoader`, unless you've thought of a fututre use for them. "followsColorTheme" can be changed to "followsThemeColor" too.

Some KDE icons has extra schemes. We could decide about them later. This is readey for merge after some cleanup and, perhaps, some tests by others. However, I emphasize the importance of patching Panel and other LXQt components that accept stylesheets before merging this. I'm sure you know why now.

tsujan:
As for extra SVG schemes, it may be good to discuss them here later.

palinek:
> For example, m_followsColorTheme and its related functions seem redundant in XdgIconLoader, unless you've thought of a fututre use for them.

That flag is to make XdgIconLoader "configurable"... it may be used standalone in other projects that don't want/need this processing. And I also intend to make this configurable in LXQt, so user can choose if she/he needs better performance or nice look.

tsujan:
Although I don't think we'll need it for better performance (no one uses Commodore 64 nowadays), we might be able to make Panel (and...) use it automatically and ignore SVG stylesheets *when it has its own stylesheet*. In that way, the user could choose an icon set suitable for Panel (for example, dark breeze for dark panels) while other apps will show the correct colors of symbolic icons.

tsujan:
The following lines in `Kvantum.cpp`, if I correctly remember:
````
  if (enoughContrast(txtCol, QApplication::palette().color(QPalette::WindowText)))
    icnMode = QIcon::Selected;
````

palinek:
Reworded & squashed ... now this should be ready to merge.

@luis-pereira can you have a look on this?

palinek:
> we might be able to make Panel (and...) use it automatically and ignore SVG stylesheets when it has its own stylesheet.

Maybe a dumb question (I'm lacking the knowledge in this area), but the style engine (kvantum, breeze...) is completely bypassed if there is an stylesheet attached for widget/application?

tsujan:
> the style engine (kvantum, breeze...) is completely bypassed if there is an stylesheet attached for widget/application?

Not necessarily. It depends on the stylesheet. The style engine isn't consulted about those parameters that are set by the stylesheet. For example, if the stylesheet sets the border and background of a push button, `QStyle::drawControl(QStyle::CE_PushButtonBevel,option,painter,widget)` won't be called but `QStyle::drawControl(QStyle::CE_PushButtonLabel,option,painter,widget)` will be called and that could cause problems if a stylesheet is incomplete.

Here, the problem is that, for Panel, `qApp->palette()` gives the palette that's used by all apps, not the stylesheet colors. So, for example, `qApp->palette().windowText().color()` may not have enough contrast with the background color set by the stylesheet. Therefore, either the app palette should be changed or Panel should ignore SVG stylesheete.

tsujan:
BTW, as I mentioned earlier, we might have to consider other SVG stylesheets too. So, I don't think this is ready for a merge yet.

See `window-close.svg` in the "16" folder of breeze! It uses `ColorScheme-NegativeText`, which has meaning only for KDE. It's better to include such cases too. There isn't any other case as far as I know.

tsujan:
I add this in case you complain about exceptions (since I won't be at my computer today).

Styling codes can't be straightforward. Exceptions should be taken into account by them (in a systematic way). That's why many good developers make ridiculous GUI's.

KDE's way of dealing with symbolic icons isn't so efficient either. They've acknowledged only the Breeze style (while QtCurve is also in "projects.kde.org"). I'm afraid this isn't unlike the attitude of Gnome devs. We could go with them as far as their code isn't too Gnomish.

palinek:
> See window-close.svg in the "16" folder of breeze! It uses ColorScheme-NegativeText, which has meaning only for KDE. It's better to include such cases too. There isn't any other case as far as I know.

Why don't we include all the classes as in [KDE iconloader](https://github.com/KDE/kiconthemes/blob/master/src/kiconloader.cpp#L77)?

tsujan:
> Why don't we include all the classes as in KDE iconloader?

You could try to add all of them. I was selective because most of them weren't needed by LXQt.

agaida:
blocked by #116 for now - patch dosn't apply clean so i could not build with it unchanged

tsujan:
> blocked by #116 for now

Whichever is merged sooner, the other one should be rebased. I use both without problem.

agaida:
right - but we should give @luis-pereira time to catch up

tsujan:
Of course. No haste.

luis-pereira:
It's fine. My only concern is performance. Does somebody have an idea on how it performs ? Are there any numbers ?

agaida:
@palinek - please rebase

palinek:
rebased

PR #123, opened 2017-04-04
xdgiconloader: Consider all existing files/images

### LINKED ISSUE #122: Wrong icons for sizes
Consider an icon theme whose icons are all SVG and whose index contains these keys:
````
...
[places/16]
Size=16
Context=Places
Type=Fixed

[places/22]
Size=22
Context=Places
Type=Fixed

[places/scalable]
Size=96
Context=Places
Type=Scalable
MinSize=16
MaxSize=512
...

````
With this arrangement, KDE correctly shows the folder icon from `places/scalable` for folders ≥ 32px while LXQt gets the folder icon from `places/16` for all sizes ≥ 32px.

palinek:
We can't assume, that the first found scalable entry is the only and
best one and stop searching.

This was an false (over)optimization or a failed merge of upstream Qt
patches.

fixes #122


tsujan:
Yes, `break` was the cause. GTM.

PR #124, opened 2017-04-06
It adds loadIcon() timing measurements.

luis-pereira:
It adds an per icon load time and a Total loadIcon() timing info.


agaida:
I guess that should only be temporary? If not i would suggest to make the debugging thing optional.

luis-pereira:
@agaida Timing is a important issue here. It's permanent. `qtxdg-iconfinder` is in it's essence but I can make optional.

palinek:
IMO it doesn't need to be optional. The qtxdg-iconfinder is developer/debug tool.

agaida:
so i'm ok with

PR #130, opened 2017-04-27
xdgiconloader: Fix XdgIconLoaderEngine::actualSize()

palinek:
Return an empty size if no suitable entry found to avoid mismatch
with the returned `pixmap()`'s size (the `QIconEngine::actualSize()` returns
the originally requested size).

closes lxde/lxqt#1282

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -672,7 +672,7 @@ QSize XdgIconLoaderEngine::actualSize(const QSize &size, QIcon::Mode mode,
             return QSize(result, result);
         }
     }
-    return QIconEngine::actualSize(size, mode, state);
+    return {0, 0};
 }
 
 QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state)


agaida:
GTM

tsujan:
@palinek 
Returning an empty size may have unforeseen consequences. Could you please be more specific about the problem this is supposed to fix? If it's about https://github.com/lxde/lxqt/issues/1282, how can the issue be reproduced?

palinek:
Here is the example:
```c++
Widget::Widget(QFrame * parent, Qt::WindowFlags f)
    : QFrame{parent, f}
{
    setFixedSize({300, 300});
    QListWidget * list = new QListWidget{};
    QString text{"dummy text"};
    for (auto i = 0; i < 10; ++i)
    {   
        text += ' ';
        text += QString::number(i);
        QListWidgetItem * item = new QListWidgetItem{QIcon::fromTheme(QStringLiteral("nonexistent")), text};
        list->addItem(item);
    }
    {   
        QScopedPointer<QAbstractItemDelegate> old_delegate{list->itemDelegate()};
        list->setItemDelegate(new QItemDelegate{});
    }
    list->setViewMode(QListView::IconMode);
    list->setIconSize({64, 64});
    QGridLayout * l = new QGridLayout;
    setLayout(l);
    l->addWidget(list, 0, 0);
}
```
and showing this widget does:
screenshot

... with this patch:
screenshot

tsujan:
OK. I'm still worried about probable side effects but I guess you've wrestled enough with this problem to know there isn't any.

tsujan:
After looking into qtbase → gui → image, I think this is all right. So, with a high probability, GTM.

palinek:
This got the [+2 also in Qt](https://codereview.qt-project.org/#/c/192904/)... merging

PR #133, opened 2017-06-12
Avoid Qt special keywords collision

luis-pereira:
Use Q_ macros instead of signals, slots, emit, foreach... to increase
compatibility with other libraries (Boost, GLib, etc).


palinek:
GTM...

just note, that in the test/qtxdg_test.cpp there is still `foreach` used

luis-pereira:
@palinek Thanks. Missed that one. It will no be compiled with other toolkits but it's good practice.
Updated.

PR #138, opened 2017-07-26
Bump Major to 3

agaida:
@luis-pereira: After Q_FOREACH libqtxdg could be considered as release ready i guess?


luis-pereira:
@agaida Yeap.

PR #141, opened 2017-09-30
Use new mimeapps.list directories

iDarkTemplar:
According to latest spec [https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html](https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html) (currently v1.0.1), there is a new list of directories to look for mimeapps.list file.

Also fix reading and writing mimeapps.list files to allow properly writing elements containing multiple values (and reading it back).


palinek:
GTM

PR #144, opened 2018-01-05
Fix symbolic SVG icons with Qt5.10

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1425 by using less private stuff.

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -48,6 +48,7 @@
 #include <QtGui/QPainter>
 #include <QImageReader>
 #include <QXmlStreamReader>
+#include <QBuffer>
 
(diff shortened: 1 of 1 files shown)


agaida:
GTM

palinek:
Even if this can work... I believe this isn't the thing we want to do, because the image is read and generated each time the `pixmap` is needed (and for each size and mode).. no caching (this was addressed by https://github.com/lxde/libqtxdg/pull/121)...

Although I would still prefer to use the QtSvg icon engine, we can avoid using it by directly using the `QSvgRenderer` and implement the caching by own.

palinek:
> Although I would still prefer to use the QtSvg icon engine, we can avoid using it by directly using the QSvgRenderer and implement the caching by own.

...actually we can use the `QSvgIconEngine` w/o the need to touch private APIs.... made a PR -> #145

PR #145, opened 2018-01-08
xdgiconloader: Cache ScalableFollowsColorEntry properly

palinek:
..by using the QSvgIconEngine under the hood.

This was done previously in be58584d0e6c11feaa6929099885ff2a7d882401,
but was reverted by 81e6ffa52ff2ec6d6bcff7e936b6a7fe04c5bbdb (this
completely lacks any caching and does read/process file and generate
image for each pixmap() call).

In this commit we don't create the (private) QSvgIconEngine manually,
but use the operators <<, >> to achieve it. Thus reducing the chance
to break things with future Qt versions (the (de)serialization will
never change for requested QDataStream::Version).

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -48,7 +48,6 @@
 #include <QtGui/QPainter>
 #include <QImageReader>
 #include <QXmlStreamReader>
-#include <QBuffer>
 
 #ifdef Q_DEAD_CODE_FROM_QT4_MAC
 #include <private/qt_cocoa_helpers_mac_p.h>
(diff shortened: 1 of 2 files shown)


palinek:
@tsujan @yan12125 can you check this, please?

tsujan:
Breaks everything. Current source (good):



This patch:

palinek:
Bah... most probably, this is the reason: https://github.com/qt/qtsvg/commit/4a9470e55ce08752501868c60762268fdf2c44af#diff-32731eae35d94dbc089ea6c66b6d085e

.

palinek:
FYI .. https://codereview.qt-project.org/#/c/216086/

tsujan:
Your latest patch works fine.

Sorry that I couldn't be of any help, having been very busy with other programs. Although I couldn't follow your reasoning, I hope you've reached your goal.

palinek:
@tsujan can you, please, check again :)...
(after studying the QSvgIconEngine, we can use just one QIcon, which holds the "svg buffer" for each QIcon::Mode)

luis-pereira:
GTM. Works as expected.

tsujan:
@palinek Your last change also works fine.

Practically, I see no difference in CPU usage (low) when changing the selected menu-item very fast, in a menu with symbolic icons (see the screenshot). However, there may be a difference in old computers.

menu

tsujan:
Oh, no! This shouldn't be merged in this state. See the toggled toolbar button:

wrong

There are delicate things that show up only with Kvantum themes ;)

tsujan:
The same toggled toolbar button without the patch (and with the current source):

right

Please note that this is a light-and-dark theme -- the toolbar is dark but the window is light -- and because of that, it can detect any problem in SVG symbolic icons.

luis-pereira:
hummm @tsujan I just checked the example you gave.  You probably have a fast computer. In laptops with say Core2  processors it makes a difference.

tsujan:
@luis-pereira Yes, my laptop is fast but I tried on a 9-year-old old computer (Core2) too. I should have said, "However, there may be a difference with some computers".

Anyhow, it's wrong in its current state.

palinek:
> shouldn't be merged in this state. See the toggled toolbar button:

@tsujan please, provide the steps to reproduce (application, icon theme, kvantum theme....), that I can debug the situation.

tsujan:
> @tsujan please, provide the steps to reproduce

There may be a misunderstanding here. You patch did work fine *before* the last commit. However, if you insist on improving that commit:

(1) Install and use the latest git Kvantum.
(2) Choose a theme like `KvSimplicityTurquoise` from Kvantum Manager.
(3) Choose an SVG icon theme like Breeze or [Plataro](https://github.com/tsujan/Plataro) (with small symbolic icons whose colors are dark by default).

Then, check the toggled toolbar buttons. You could also choose a theme like `KvCurvesLight` to see what happens with selected menu-items but they're OK with that commit.

palinek:
Hopefully now it is correct...

@tsujan please, confirm.

tsujan:
@palinek This time, it seems OK -- or, at least, I couldn't find a problem.

palinek:
> This time, it seems OK -- or, at least, I couldn't find a problem.

Thanks. Squashed.... I'll wait few days before merging to let you/others find potential bugs.

tsujan:
> I'll wait few days before merging to let you/others find potential bugs.

That's a wise decision.

tsujan:
5 days passed and this is definitely OK; please merge it! The next release shouldn't be without it.

tsujan:
Just FYI:

@palinek's patch may have prevented a very different issue that exists under KDE. A link: https://github.com/tsujan/Kvantum/issues/182#issuecomment-368459076

The issue in question appeared in disabled icons when colorizing was manipulated by Kvantum for dark-and-light themes. However, it doesn't exist under LXQt, perhaps, because of @palinek's patch (just a guess).

Anyhow, LXQt's code looks better than KDE's.

PR #149, opened 2018-01-15
xdgiconloader: Add GTK cache revalidation

palinek:
In case an icon is added and the GTK cache is refreshed the already
running Qt application does not get the information from GTK cache.
That's because the cache file is changed (modified or deleted+newly
created), but the mapped memory used to read the file is not affected.

We will re-map the cache file for such situations to properly use the
cache also after the content change.

### DIFF
--- a/xdgiconloader/xdgiconloader.cpp
+++ b/xdgiconloader/xdgiconloader.cpp
@@ -117,7 +117,6 @@ private:
     const unsigned char *m_data;
     quint64 m_size;
     bool m_isValid;
-    QFileSystemWatcher m_watcher;
 
     quint16 read16(uint offset)
     {
@@ -137,6 +136,7 @@ private:
             | m_data[offset+1] << 16 | m_data[offset] << 24;
     }
 };
+Q_GLOBAL_STATIC(QFileSystemWatcher, gtkCachesWatcher);
 
 
 QIconCacheGtkReader::QIconCacheGtkReader(const QString &dirName)
@@ -148,7 +148,16 @@ QIconCacheGtkReader::QIconCacheGtkReader(const QString &dirName)
     // Note: The cache file can be (IS) removed and newly created during the
     // cache update. But we hold open file descriptor for the "old" removed
     // file. So we need to watch the changes and reopen/remap the file.
-    QObject::connect(&m_watcher, &QFileSystemWatcher::fileChanged, &m_watcher, [this] { m_isValid = false; });
+    QObject::connect(static_cast<QFileSystemWatcher *>(gtkCachesWatcher), &QFileSystemWatcher::fileChanged
+            , static_cast<QFileSystemWatcher *>(gtkCachesWatcher), [this](const QString & path)
+        {
+            if (m_file.fileName() == path)
+            {
+                m_isValid = false;
+                // invalidate icons to reload them ...
+                QIconLoader::instance()->invalidateKey();
+            }
+        });
     reValid(false);
 }
 
@@ -168,8 +177,8 @@ bool QIconCacheGtkReader::reValid(bool infoRefresh)
 
     // Note: If the file is removed, it is also silently removed from watched
     // paths in QFileSystemWatcher.
-    if (!m_watcher.files().contains(m_cacheFileInfo.absoluteFilePath()))
-        m_watcher.addPath(m_cacheFileInfo.absoluteFilePath());
+    if (!gtkCachesWatcher->files().contains(m_cacheFileInfo.absoluteFilePath()))
+        gtkCachesWatcher->addPath(m_cacheFileInfo.absoluteFilePath());
 
     if (!m_file.open(QFile::ReadOnly))
         return m_isValid;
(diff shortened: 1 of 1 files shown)


tsujan:
I tested this with Retext while `/usr/share/applications` was open in pcmanfm-qt before installation. The icon of Retext was "unknown". Then I removed Retext, reinstalled the current `libqtxdg`, logged out and in, opened  `/usr/share/applications` and installed Retext again: the icon was shown correctly this time.

palinek:
> I tested this with Retext while /usr/share/applications was open in pcmanfm-qt before installation. The icon of Retext was "unknown".

That's because `pcfmanfm-qt` is faster than the installation process (extract package files, trigger icon cache update...)

tsujan:
@palinek Currently I'm very busy with another problem in another app (not belonging to LXQt), so I can't look into the theory but may only be able to test (having a ready PKGBUILD). For whatever reason, this doesn't work, while the current code works.

palinek:
While trying to find out the reason for not refreshing the icon after the gtk cache update I realized, that the problem is connected with the libfm-qt and its icon handling.... 
To check this...also the lxde/libfm-qt#165 (and lxde/pcmanfm-qt#642 is needed).

tsujan:
I honestly believe that `QFileSystemWatcher` is a bad overkill here.

palinek:
> I honestly believe that QFileSystemWatcher is a bad overkill here.

Why? ... https://stackoverflow.com/questions/535768/what-is-a-reasonable-amount-of-inotify-watches-with-linux

tsujan:
> Why?....

Strangely, our roles -- yours and mine -- are reversed in this case ;)

agaida:
@tsujan - to be true - i don't see prolems (overkill) if using QFilesystemWatcher - which problems do you expect?

tsujan:
@agaida My opinion won't change on this ;) Using a *continuous* "watcher" to solve this problem is like using a modern computer just to do the job of a calculator. And Qt doesn't lie in this case; quoting from http://doc.qt.io/qt-5/qfilesystemwatcher.html: "The act of monitoring files and directories for modifications consumes system resources..."

agaida:
Instead of opinions - wouldn't it be nice to have some benchmarks about the real costs of a continous watcher? I know that the watcher have some costs, but please - i only want to know how much :D

tsujan:
> wouldn't it be nice to have some benchmarks

I didn't find any on the Internet. It's only said that it's the best on Linux -- compared to Mac and Windows -- because it uses inotify. I have a plan to try it in FeatherPad.

Here, my argument (or opinion) is that using a watcher instead of only checking missing icons -- as is already done -- doesn't seem like a nice method; it's out of proportion with its result, which is just a workaround for a drawback of using the gtk icon cache.

Either way, the gtk icon cache has proved not to be the best solution it was advertised to be.

agaida:
what about to make the watcher optional - and just see what happend?

tsujan:
> and just see what happend?

The current code of `libqtxdg` shows the newly installed icons correctly. This PR should do the same thing. So, nothing special would happen ;) In this PR, the line
```c++
if (!result.isEmpty()) {
```
is reversed to
```c++
if (cache->isValid()) {
```
and, instead, lots of lines are added for `QFilesystemWatcher` but the result is the same.

This isn't about the result but about the elegance, IMO.

tsujan:
In other words, I don't think you'll see any difference on old machines either -- but if you did, please tell me!

tsujan:
To be super transparent and to save myself from adding further comments on this PR:

There's no mystery in this: Choosing the gtk icon cache was a mistake from start, exactly because Qt couldn't change it -- some method like that of KDE should have been adapted instead.

This PR tries to save that mistake by setting up a *permanent* file system watcher instead of searching for missing icons *only when they're needed* (= the current code). That looks irrational to me, even if I suppose that `QFilesystemWatcher` is very resource-friendly -- contrary to what the Qt doc says.

agaida:
after running that a few days i can see no downside, even if it might be not the best of all solutions - so i will be open for any impovement in the future.

tsujan:
> after running that a few days i can see no downside

Had you seen any downside *without it*?

agaida:
Having an refreshed icon cache is fine - it might be that the permanent watcher is really overkill. But as long he don't hurt ... I just have no clue, how to solve this problem right. Anyways - now that it is solved, we can talk again if it was the best way to solve it.

It is sometimes hard for me to make that kind of decisions - so please - if possible, discuss it and find the right one. I was fine with both. This is also true with the damn sudo thing - in the end we need a decision, i'm not affraid to decide, but the chance is high, that the decision will not be optimal.

tsujan:
@agaida I always accept your decisions about LXQt -- as far as I know, they've been the best -- but just wanted to know if you saw a real problem before this because I don't.

As for a refreshed icon cache, it isn't under Qt's control. The cache was always updated correctly but wasn't seen by Qt. This PR works around that by watching it.

palinek:
@agaida rhe 'fixup...' commits in my PRs are there just for reviewers to better see the progress and changes made. Please, next time don't merge them directly into master, they are intended to be melded into the previous commit.

PR #150, opened 2018-03-09
Use Qt5 new signal slot syntax

luis-pereira:


tsujan:
Since Qt-5.7, `QOverload` exists.

luis-pereira:
libqtxdg Qt requirement is Qt-5.7. On purpose.
We are not the only ones using it. The deepin desktop also uses it. https://github.com/linuxdeepin/qt5integration
IMO this tiny thing isn't a reason to require Qt 5.7.

tsujan:
OK.

Not intending to start a discussion on such an unimportant thing, I just think it's better to continue developing LXQt without caring for specific distros with old Qt versions (< latest - 3). Although I always care for Ubuntu LTS + Debian Stable (having none of them) in my non-LXQt codes, I never saw LXQt doing so before.

agaida:
and i see a problem if we would care - LTS means long term stable - so these release are not meant to be used with current things. LTS means no changes whithin the lifetime of the release.

tsujan:
> and i see a problem if we would care

I completely agree when it comes to LXQt because LXQt is a set on interrelated programs and supporting LTS distros would be very time--consuming if not impossible. Moreover, I don't think all LXQt can be compiled against Qt < 5.7 (I may be wrong though).

I do that for my own apps (1) to get more feedback but, mainly, out of nostalgia because I used debian and ubuntu for a long time.

palinek:
GTM

luis-pereira:
@agaida @tsujan  I can't understand what you guys are talking about.
I didn't talk about any distros. The [Deepin Desktop](https://www.deepin.org/en/) uses libqtxdg. The same way we use some KDE libraries.

tsujan:
@luis-pereira For whatever reason, recently, there have been lots of misunderstanding about trivial things. [Wikipedia](https://en.wikipedia.org/wiki/Deepin) says, "Deepin is a popular Chinese Linux distribution based on Debian....". So, it *is* a distro and, if it doesn't have Qt-5.7, it should be old.

tsujan:
> The deepin desktop, is not tied to any distro...

It became somehow significant after this. A distro or not, why should we follow its developers and not conversely? Should we do so for other distros/non-distros too? My answer is "no".

Merge it to end an unwanted discussion.

luis-pereira:
@tsujan That's not the point. I don't see any technical reason to update the minimum required version. It's one line of code.
At the same time, QTerminal requiring Qt 5.7 has several cases where QOverload wasn't used.

No, I will not merge it. All PR's are equal.

agaida:
Deepin uses  libqt5xdg2 and liblxqt0 0.11.1 beside Qt 5.6.1 in their current release - so Deepin is no reason to take care.

luis-pereira:
@agaida For My Information, Which component depends on liblxqt ?

agaida:
```
$ apt rdepends  libqt5xdg2
libqt5xdg2
Reverse Depends:
  Depends: libqt5xdg-dev (= 2.0.0-6)
  Depends: dde-qt5integration (>= 1.0.0)
  Depends: screengrab (>= 1.0.0)
  Depends: lxqt-session (>= 1.3.0)
  Depends: lxqt-runner (>= 1.0.0)
  Depends: lxqt-powermanagement (>= 1.0.0)
  Depends: lxqt-panel (>= 1.0.0)
  Depends: lxqt-notificationd (>= 1.0.0)
  Depends: lxqt-config (>= 1.0.0)
  Depends: lxqt-about (>= 1.0.0)
  Depends: libqt5xdg2-dbgsym (= 2.0.0-6)
  Depends: dde-qt5integration (>= 1.0.0)
  Depends: liblxqt0 (>= 1.0.0)
deepin@Deepin:~/Desktop$ apt rdepends dde-qt5integration
dde-qt5integration
Reverse Depends:
  Breaks: qt5dxcb-plugin (<< 0.2.8.1)
  Recommends: dde-file-manager
  Breaks: qt5dxcb-plugin (<= 0.2.8.1)
  Depends: dde
  Recommends: startdde
  Depends: deepin-appstore (>= 0.2.0)
  Recommends: startdde
  Depends: dde-qt5integration-dbgsym (= 0.2.8.3-1)
  Depends: deepin-appstore (>= 0.2.0)
  Depends: dde
  Recommends: dde-file-manager
deepin@Deepin:~/Desktop$ apt rdepends dde
dde
Reverse Depends:
  Depends: deepin-desktop-environment
```

agaida:
@luis-pereira - sorry, typo
```
apt-cache rdepends liblxqt0 | sort | uniq 
liblxqt0
  liblxqt0-dbgsym
  liblxqt0-dev
  liblxqt-l10n
  lxqt-about
  lxqt-admin
  lxqt-config
  lxqt-globalkeys
  lxqt-notificationd
  lxqt-openssh-askpass
  lxqt-panel
  lxqt-policykit
  lxqt-powermanagement
  lxqt-runner
  lxqt-session
  lxqt-sudo
```

PR #155, opened 2018-05-15
xdgaction: Remove unneeded processEvents() call

palinek:
..as it can lead to unwanted recurrency and cause segfaults.

E.g.: lxqt-panel mainmenu searching:
 (0x7ffe4effb0e0) Warning: QWidget::repaint: Recursive repaint detected
 (0x7ffe4effb0e0) Warning: QWidget::paintEngine: Should no longer be called
 (0x7ffe4effb0e0) Warning: QPainter::begin: Paint device returned engine == 0, type: 1
 (0x7ffe4effb0e0) Warning: QPainter::save: Painter not active
 (0x7ffe4effb0e0) Warning: QPainter::setPen: Painter not active
 (0x7ffe4effb0e0) Warning: QPainter::drawRects: Painter not active
 (0x7ffe4effb0e0) Warning: QPainter::restore: Unbalanced save/restore

### DIFF
--- a/qtxdg/xdgaction.cpp
+++ b/qtxdg/xdgaction.cpp
@@ -118,5 +118,4 @@ void XdgAction::updateIcon()
     setIcon(mDesktopFile.icon());
     if (icon().isNull())
         setIcon(XdgIcon::fromTheme(QLatin1String("application-x-executable")));
-    QCoreApplication::processEvents();
 }


palinek:
I'm getting it with latest master (lxqt-panel, libqtxdg, liblxqt) with mainmenu configuration:
[x] filter menu entries
[x] show matching entries
[ ] hide menu entries while searching

...then typing any "trash" into search crashes the panel

tsujan:
Strange! I can't reproduce a crash with the above-mentioned settings. Moreover, a warning like "QPainter::restore: Unbalanced save/restore" may have another cause.

However, as far as I've seen, `QCoreApplication::processEvents()` is usually a bad idea. Whether it should be removed or replaced in this special case, I don't know.

tsujan:
Succeeded in making the panel crash by using the search tens of time -- lost the backtrace though. Hopefully this will be the fix; already applied it.

agaida:
couldn't reproduce the crash - no messages in log nor when running in gdb. 
Environment: Current Arch with kde-unstable activated, LXQt built wit Qt 5.11rc1

GTM

agaida:
erm - running into some issues with pcmanfm-qt - not really sure if the patch is the reason - so

Not good to merge right now, investigating.

agaida:
ok - sorry for the noise: lxqt/libfm-qt/issues/164 was filed already - GTM again

PR #160, opened 2018-07-10
Adds Qt Strings abbreviations macros

luis-pereira:
It reduces the clutter. Ex: instead of writing:
    ```if (node.hasAttribute(QStringLiteral("http-contents-length")))```
    just write:
    ```if (node.hasAttribute(QSL("http-contents-length")))```


agaida:
GTM

tsujan:
GTM

PR #162, opened 2018-08-17
Adds a MIME Applications Associations (XdgMimeApps) class and companion tools

luis-pereira:
When fully implemented it will render XdgDesktopFileCache obsolete.
Currently GLib implementation is used as backend. The design allows
other backends to be used, if desired.

Note to packagers: GLib is now a explicit dependency.


luis-pereira:
In conjunction with https://github.com/lxqt/lxqt/issues/1528  will provide the core functionality to solve https://github.com/lxqt/lxqt/issues/1513 and alikes.

This is the kickstart.

luis-pereira:
`qtxdg-mat` command line tool added. 
Currently supports set/get the default App for a mimeType.

`open` will be the next command to be added.

luis-pereira:
@ALL It's usable now, IMO.

PR #163, opened 2018-08-19
Cover all KDE colors schemes for symbolic icons

tsujan:
In our support for KDE symbolic icons, we only considered `.ColorScheme-Text`. But some Breeze symbolic icons use KDE-specific schemes like `.ColorScheme-NegativeText`. Of course, we don't use such colors but even `.ColorScheme-Text` is something we've taken from KDE (it could be any string otherwise). So, it's better to cover all KDE text colors but set all of them to the same color.

Without this patch, some Breeze icons with KDE-specific schemes (like the 16-px `window-close.svg` or `application-exit.svg`) would always be pitch black -- even with dark Qt styles -- because their schemes are erased. But with this patch, they are like other icons: light on dark backgrounds and dark on light ones.

In short, this is needed if we want KDE icon sets to be really usable under LXQt too.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -841,8 +841,8 @@ QPixmap ScalableFollowsColorEntry::pixmap(const QSize &size, QIcon::Mode mode, Q
             const QPalette pal = qApp->palette();
             // Note: indexes are assembled as in qtsvg (QSvgIconEnginePrivate::hashKey())
             QMap<int, QString> style_sheets;
-            style_sheets[(QIcon::Normal<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text{color:%1;}").arg(pal.windowText().color().name());
-            style_sheets[(QIcon::Selected<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text{color:%1;}").arg(pal.highlightedText().color().name());
+            style_sheets[(QIcon::Normal<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text,.ColorScheme-PositiveText,.ColorScheme-NeutralText,.ColorScheme-NegativeText{color:%1;}").arg(pal.windowText().color().name());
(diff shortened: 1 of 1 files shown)


src/xdgiconloader/xdgiconloader.cpp
```diff
@@ -841,8 +841,8 @@ QPixmap ScalableFollowsColorEntry::pixmap(const QSize &size, QIcon::Mode mode, Q
             const QPalette pal = qApp->palette();
             // Note: indexes are assembled as in qtsvg (QSvgIconEnginePrivate::hashKey())
             QMap<int, QString> style_sheets;
-            style_sheets[(QIcon::Normal<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text{color:%1;}").arg(pal.windowText().color().name());
-            style_sheets[(QIcon::Selected<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text{color:%1;}").arg(pal.highlightedText().color().name());
+            style_sheets[(QIcon::Normal<<4)|QIcon::Off] = QStringLiteral(".ColorScheme-Text,.ColorScheme-PositiveText,.ColorScheme-NeutralText,.ColorScheme-NegativeText{color:%1;}").arg(pal.windowText().color().name());
```
palinek: Why do we define the same color for all of the classes? Shouldn't we do it the way as kiconloader does it?
https://github.com/KDE/kiconthemes/blob/master/src/kiconloader.cpp#L75
https://github.com/KDE/kiconthemes/blob/master/src/kiconloader.cpp#L898

tsujan: Explained in the patch comment: those extra colors aren't Qt palette colors but KDE specific ones. Here, we just tolerate their names.

palinek:
>  those extra colors aren't Qt palette colors

In KDE code I see usage of 3 colors from Qt palette in different way for specific class:
```c++
        state == KIconLoader::SelectedState ? pal.highlightedText().color().name() : pal.windowText().color().name(),
        state == KIconLoader::SelectedState ? pal.highlight().color().name() : pal.window().color().name(),
        state == KIconLoader::SelectedState ? pal.highlightedText().color().name() : pal.highlight().color().name(),
```
Why don't we try follow it as much as possible?

palinek:
> Would you do that?

Done in #164

PR #164, opened 2018-08-22
xdgiconloader: Support more classes for "FollowColors" icons

palinek:
..and leave default colors for unknown/unsupported classes

Example... window-close from breeze:
- before
screenshot
- after
screenshot


tsujan:
This seems much better because the colors of unsupported schemes are preserved. Hopefully, it's CPU friendly. Will test it soon.

agaida:
looks promising

tsujan:
GTM

PR #167, opened 2018-11-19
Set cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)

agaida:
Removed the cmake version checks and just set it


examples/use-qtxdg/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 project(use-qtxdg)
-cmake_minimum_required(VERSION 2.8.12)
```
palinek: Why did you remove it here? It should (must) just be the first statement in the main CMakeLists.txt file... and this an example of project using the xdg, isn't it?

agaida: Good point, i saw it as part of this project only.

PR #170, opened 2018-12-06
Consider device pixel ratio when painting

tsujan:
Otherwise, (view item) icons will be blurry with PNG icon sets and scale factors greater than 1. `KIconEngine::paint()` does the same thing.

Also see https://github.com/lxqt/pcmanfm-qt/issues/834#issuecomment-444715167

IMHO, it's better not to merge this before the 0.14 release because it might need more tests.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -631,7 +631,8 @@ void XdgIconLoaderEngine::paint(QPainter *painter, const QRect &rect,
 #if defined(Q_DEAD_CODE_FROM_QT4_MAC)
     pixmapSize *= qt_mac_get_scalefactor();
 #endif
-    painter->drawPixmap(rect, pixmap(pixmapSize, mode, state));
+    const qreal dpr = painter->device()->devicePixelRatioF();
+    painter->drawPixmap(rect, pixmap(pixmapSize * dpr, mode, state));
 }
 
 /*


palinek:
GTM

palinek:
Just a question.. I had a quick look into qt sources...
```c++
      case QIconEngine::ScaledPixmapHook:
          {
              QIconEngine::ScaledPixmapArgument &arg = *reinterpret_cast<QIconEngine::ScaledPixmapArgument*>(data);
              // QIcon::pixmap() multiplies size by the device pixel ratio.
              const int integerScale = qCeil(arg.scale);
              QIconLoaderEngineEntry *entry = entryForSize(m_info, arg.size / integerScale, integerScale);
              arg.pixmap = entry ? entry->pixmap(arg.size, arg.mode, arg.state) : QPixmap();
          }
          break;
```
and
```c++
  QPixmap QIcon::pixmap(QWindow *window, const QSize &size, Mode mode, State state) const
  {
      if (!d)
          return QPixmap();
  
      qreal devicePixelRatio = qt_effective_device_pixel_ratio(window);
  
      // Handle the simple normal-dpi case:
      if (!(devicePixelRatio > 1.0)) {
          QPixmap pixmap = d->engine->pixmap(size, mode, state);
          pixmap.setDevicePixelRatio(1.0);
          return pixmap;
      }
  
      // Try get a pixmap that is big enough to be displayed at device pixel resolution.
      QIconEngine::ScaledPixmapArgument scalePixmapArg = { size * devicePixelRatio, mode, state, devicePixelRatio, QPixmap() };
      d->engine->virtual_hook(QIconEngine::ScaledPixmapHook, reinterpret_cast<void*>(&scalePixmapArg));
      scalePixmapArg.pixmap.setDevicePixelRatio(d->pixmapDevicePixelRatio(devicePixelRatio, size, scalePixmapArg.pixmap.size()));
      return scalePixmapArg.pixmap;
  }
```

is the handling wrong there?

tsujan:
Yes, I also saw that. The ways Qt, KDE and LXQt do it are different from each other; I think that we should consider the differences in other functions too. However, I remember that scaling was terrible when it was first introduced in Qt.

I'm pretty sure that our implementation has yet another issue: the scaled icons aren't always drawn as smooth as they can be, especially when they are SVG. Perhaps, somewhere, we get a too big pixmap and then make it small again. This problem can't be seen easily.

tsujan:
@agaida Contrary to what I said in the PR comment, this can be merged before the next release because (1) @palinek approved of it, and (2) it passed all my tests.

Other issue(s) can wait.

PR #182, opened 2019-04-09
xdgdesktopfile: Avoid freeze for DBusActivatable apps

### LINKED ISSUE #181: launching app, crashes lxqt-panel
Attempting to launch a program crashes the panel and whole session, if:
1. There is already a running (but frozen / hanging / crashed) instance
2. In the desktop file it is marked as DBus Activatable

How to reproduce:
1. Use lxqt
2. Open terminal 
3. Run e.g gnome-maps (or a program which has DBusActivatable=true in its desktop file)
4. Hit ctrl + z
5. Try to open gnome-maps from lxqt-panel-menu
6. Everything freezes
(Recent DBusActivatable version required of course)

Changing the value in the desktop file from
DBusActivatable=true
to
DBusActivatable=false

PS: if the program is dbus activatable and just has a slow startup the ui freezes until it starts

confirms this is the issue (the session and panel do not freeze with DBusActivatable=false)

palinek:
For case when the DBusActivatable application is unresponsive the
startDetached() can block for a long time (the Qt default 25s DBus
timeout). Blocking can't be avoided while using QDBusInterface, because
the object can block directly in its contructor.

So we use the generated interface class and set the timeout for the DBus
call to 1500ms (can be overriden env variable QTXDG_DBUSACTIVATE_TIMEOUT).

See also: https://bugreports.qt.io/browse/QTBUG-75016

fixes #181

### DIFF
--- a/src/qtxdg/CMakeLists.txt
+++ b/src/qtxdg/CMakeLists.txt
@@ -1,3 +1,5 @@
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
 set(QTX_LIBRARIES Qt5::Widgets Qt5::Xml Qt5::DBus)
 
 set(libqtxdg_PUBLIC_H_FILES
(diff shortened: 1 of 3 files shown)


palinek:
@luis-pereira please, take a look into the cmake part

damianatorrpm:
Works amazing. Panel freezes only for QTXDG_DBUSACTIVATE_TIMEOUT amount of time and doesn't crash, in the background 25 sec later

`error sending Activate message to application: Timeout was reached
`

agaida:
GTM

agaida:
unblocked for debian buster

PR #183, opened 2019-04-09
Modernize

luis-pereira:
Use C++11 features.
Done with `clang-tidy`.

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -169,7 +169,7 @@ QString &escapeExec(QString& str)
 QString &doUnEscape(QString& str, const QHash<QChar,QChar> &repl)
 {
     int n = 0;
-    while (1)
+    while (true)
     {
         n=str.indexOf(QLatin1String("\\"), n);
         if (n < 0 || n > str.length() - 2)
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -31,7 +31,7 @@
 #include "xdgdirs.h"
 #include "xdgicon.h"
 
-#include <stdlib.h>
+#include <cstdlib>
 #include <unistd.h>
 
 #include <QDebug>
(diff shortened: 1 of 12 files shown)


src/qtxdg/xdgdesktopfile.cpp
```diff
@@ -986,7 +986,7 @@ static QStringList parseCombinedArgString(const QString &program)
 void replaceVar(QString &str, const QString &varName, const QString &after)
 {
     str.replace(QRegExp(QString::fromLatin1("\\$%1(?!\\w)").arg(varName)), after);
-    str.replace(QRegExp(QString::fromLatin1("\\$\\{%1\\}").arg(varName)), after);
+    str.replace(QRegExp(QString::fromLatin1(R"(\$\{%1\})").arg(varName)), after);
```
palinek: ...and why not in the line above? :)

tsujan:
Nice!

(1) Raw string literals have problem with translation updates with Qt < 5.12.2 (I think @palinek fixed them in Qt). Do we require that version?
(2) Now that the code is updated, why not replacing `QRegExp` with `QRegularExpression` too?

palinek:
> (1) Raw string literals have problem with translation updates with Qt < 5.12.2 (I think @palinek fixed them in Qt). Do we require that version?

It seems, it is backported into versions from 5.12 -> https://github.com/qt/qttools/commit/ee9f6ec9b9e8f0d18ad7e8fededb5292abb91912

tsujan:
> It seems, it is backported into versions from 5.12

I recompiled it with your patch since Qt 5.11 (needed it for my apps) but, only when 5.12.2 was released, I didn't need a recompilation.

Anyhow, it'll be OK if there's nothing to translate in files with raw string literals.

luis-pereira:
> Anyhow, it'll be OK if there's nothing to translate in files with raw string literals.

That's the case.

tsujan:
> That's the case.

Good.

src/qtxdg/xdgdesktopfile.cpp
luis-pereira: clang-tidy, probably, got confused by the parentheses.
I only checked what it changed.

``` cpp
R"regexp(\$%1(?!\w))regexp"
```
cuts it.

Will add ASAP.

PR #184, opened 2019-04-11
Drop QRegExp.

luis-pereira:
QRegularExpression is recommended over QRegExp.
QRegularExpression has Perl compatible syntax and performs a lot better.

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -48,6 +48,7 @@
 #include <QMimeDatabase>
 #include <QMimeType>
 #include <QProcess>
+#include <QRegularExpression>
 #include <QSettings>
 #include <QStandardPaths>
 #include <QTextStream>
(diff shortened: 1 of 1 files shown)


tsujan:
GTM.

Do you know where the header of `QRegExp` was included?

tsujan:
BTW and if `QRegExp` is going to be replaced in other places too:

`QRegularExpression` is surprisingly perl compatible. The only issue I've found in it is that it doesn't understand `{,n}`; it should be changed to `{0,n}`. I'm not sure if this is a bug.

luis-pereira:
> GTM.
> 
> Do you know where the header of `QRegExp` was included?

`QString`.

luis-pereira:
> BTW and if `QRegExp` is going to be replaced in other places too:
> 
> `QRegularExpression` is surprisingly perl compatible. The only issue I've found in it is that it doesn't understand `{,n}`; it should be changed to `{0,n}`. I'm not sure if this is a bug.

I think it was designed to be perl compatible.
Never faced with that case. But.... `QRegExp` has a lot of quirks.

And yes, we should drop `QRegExp` everywhere.

tsujan:
> QString.

Thanks!

> QRegExp has a lot of quirks.

It was a torture when used seriously.

PR #198, opened 2019-12-17
Add XdgDesktopFile::actionExpandExecString

damianatorrpm:
Same as expandExecString but for actions.

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -924,6 +924,11 @@ QString XdgDesktopFile::actionName(const QString & action) const
         : QString{};
 }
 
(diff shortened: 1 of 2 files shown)


damianatorrpm:
I don't see this options in github GUI. I have never used git in-depth except github and do not intend to do so. I can send the patch via e-mail alternatively.

luis-pereira:
@damianatorrpm 
git is not build system. It's a distributed control version system. The most widely used these days. LXQt  build system is CMake.
LXQt uses git. So a developer must have a minimal understanding on how to use it.
You can use gitk to do a mixed reset to the first commit, them use git gui to amend the commit. Then push as usual, but use the -f flag.

tsujan:
I also had problems with git when I started to use it several years ago. But, after a few weeks, everything became clear. IMHO, the best way of using git is command line; git front-ends/GUIs hide git.

PR #199, opened 2019-12-27
Fix XdgMimeType with Qt 5.14

yan12125:
Reverts lxqt/libqtxdg#197

Breaks building of lxqt-panel and lxqt-config with Qt 5.14 and GCC 9.2 on Arch Linux.

### DIFF
--- a/src/qtxdg/xdgmimetype.h
+++ b/src/qtxdg/xdgmimetype.h
@@ -29,8 +29,6 @@
 
 #include <QDebug>
 
-#include <algorithm>
-
 class XdgMimeTypePrivate;
 
 //! Describes types of file or data, represented by a MIME type string.
@@ -85,7 +83,8 @@ public:
     void swap(XdgMimeType &other)
     {
         QMimeType::swap(other);
-        std::swap(dx, other.dx);
+        // Not using std::swap here as that does not work with Qt 5.14
+        dx.swap(other.dx);
     }
 
     //! Destructs the mimetype


yan12125:
This commit https://github.com/qt/qtbase/commit/2ab3af564c7c8424b558c5902614d1a410850acf, which is included in Qt 5.14, seems the key change.

tsujan:
`qSwap` is marked as "obsolete" (https://doc.qt.io/qt-5/qtalgorithms-obsolete.html) and yet, we can't replace it because of that Qt commit? Or, maybe, there's another way?

luis-pereira:
@tsujan The updated code is in libqtxdg and it builds fine. 
Then the compilation fails in moc/ui generated code. I have some ideas of what might be triggering it.
Until Jan 7, I won't be able to test my hypotheses.

@yan12125 Can you add `#include <utility>` next to `#include<algorithm>` in libqtxdg `xdgmimetype.h` and check for any changed ?

yan12125:
> Can you add #include <utility> next to #include<algorithm> in libqtxdg xdgmimetype.h and check for any changed ?

The error remains.

> Or, maybe, there's another way?

I managed to get it work on Qt 5.14 without reverting to deprecated `qSwap()`. The commit is force-pushed to this pull request.

tsujan:
> I managed to get it work on Qt 5.14 without reverting to deprecated qSwap().

Very nice!

yan12125:
Gonna merge as I'm relatively confident that this approach is equivalent to the old qSwap one and this has been broken for some time.

PR #200, opened 2020-01-11
Use std::swap()

luis-pereira:
Commit 6f597b30cef2bdefc9ca5362cdf195294ebb9556 replaced qSwap() with
std::swap(). It caused a FTBFS with Qt5.14
Commit c6b89c7af127d49739c9d2d9d323c931c509dcaa fixed it.
But it can be made to work with std::swap(), which is more readable, and
resilient to changes in Qt internals.
Tested with Qt5.14.2 and Qt5.14.1

### DIFF
--- a/src/qtxdg/xdgmimetype.cpp
+++ b/src/qtxdg/xdgmimetype.cpp
@@ -22,6 +22,7 @@
 
 #include "xdgicon.h"
 
+#include <algorithm>
 
 class XdgMimeTypePrivate : public QSharedData {
 public:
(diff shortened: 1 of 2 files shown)


luis-pereira:
@yan12125 The `noexcept` isn't the reason why it fixes the issue. Moving the `swap` implementation to the cpp file is the reason. `XdgMimeTypePrivate` is forward declared in the .h file, so, when included by other code... leads to a incomplete type.

Added the `noexcept` to match [`std::swap()` definition](https://en.cppreference.com/w/cpp/algorithm/swap).

Comments about what  ?

yan12125:
> Moving the swap implementation to the cpp file is the reason. XdgMimeTypePrivate is forward declared in the .h file, so, when included by other code... leads to a incomplete type.

Aha, got it. I was just worrying that future readers of this commit may not get the point of the change. With this explanation, no further comment is needed IMO.

luis-pereira:
Ok. So let's merge it.

PR #202, opened 2020-02-21
Avoid hard-coded paths

voidpin:


luis-pereira:
@voidpin 
The hard-coded paths you refer are defined in the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). We are just following specification.

The hard-coded values are fallback values used only when the corresponding environment variables are not set. Fallback values are immutable.  See [here](https://github.com/lxqt/libqtxdg/blob/master/src/qtxdg/xdgdirs.cpp#L263) and [here](https://github.com/lxqt/libqtxdg/blob/master/src/qtxdg/xdgdirs.cpp#L288).

`${PREFIX}` is defined in the `CMakeLists.txt`. It's value is not resolved in the C++ code.  The resulting dir will be literally be: `${PREFIX}/share`.

The correct way to change the values is to set the  `XDG_DATA_DIRS` and `XDG_CONFIG_DIRS` environment variables.

PR #203, opened 2020-02-24
Sharp SVG icons with scale factors > 1

tsujan:
Fixes https://github.com/lxqt/libqtxdg/issues/178

The SVG icon engine gives a too big pixmap with scale factors > 1, so that the final drawing will not be sharp if the real size is used. This patch scales down the pixmap to the actual size in order to get a sharp SVG icon, especially with non-integer scale factors.

Now, for the first time in LXQt, I get sharp SVG icons with `QT_SCALE_FACTOR=1.5`:) Before the patch:

before

After the patch:

after

NOTE: There's a comment in the patch that says, "The cause of this old behavior is unknown to me (@tsujan)." If you know the cause, please replace the comment!

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -623,7 +623,7 @@ void XdgIconLoaderEngine::paint(QPainter *painter, const QRect &rect,
 {
     QSize pixmapSize = rect.size();
     const qreal dpr = painter->device()->devicePixelRatioF();
-    painter->drawPixmap(rect, pixmap(pixmapSize * dpr, mode, state));
+    painter->drawPixmap(rect, pixmap(QSizeF(pixmapSize * dpr).toSize(), mode, state));
 }
 
 /*
@@ -815,7 +815,17 @@ QPixmap ScalableEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State 
         svgIcon = QIcon(filename);
 
     // Simply reuse svg icon engine
-    return svgIcon.pixmap(size, mode, state);
+    QPixmap pm = svgIcon.pixmap(size, mode, state);
+    // WARNING: The SVG icon engine gives a too big pixmap with scale factors > 1,
+    // so that the final drawing will not be sharp if the pixmap is not scaled here.
+    // The cause of this old behavior is unknown to me (@tsujan).
+    if (!pm.isNull())
+    {
+        const auto actualSize = svgIcon.actualSize(size, mode, state);
+        if (actualSize != pm.size())
+            pm = pm.scaled(actualSize, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
+    }
+    return pm;
 }
 
 static const QString STYLE = QStringLiteral("\n.ColorScheme-Text, .ColorScheme-NeutralText {color:%1;}\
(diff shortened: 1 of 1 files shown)


src/xdgiconloader/xdgiconloader.cpp
```diff
@@ -815,7 +815,10 @@ QPixmap ScalableEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State
         svgIcon = QIcon(filename);
 
     // Simply reuse svg icon engine
-    return svgIcon.pixmap(size, mode, state);
+    // WARNING: SVG icon engine gives a too big pixmap with scale factors > 1,
+    // so that the final pixmap will not be sharp if the real size is used here.
+    // The cause of this old behavior is unknown to me (@tsujan).
+    return svgIcon.pixmap(QSizeF(size / qApp->devicePixelRatio()).toSize(), mode, state);
```
guoyunhe: `qApp->devicePixelRatio()` should be avoided because Wayland supports different scale factors on multiple display set up (x2 for laptop screen, x1.5 for external screen). The DPR from widget is preferred. The `parent` widget should be passed to this icon loader class and it will use `parent->devicePixelRatioF()`

tsujan: > The parent widget should be passed

Widgets (or their parents) don't have meaning on this level.

If wayland supports different scale factors, they should be reported to the app and the app show know about them.

guoyunhe: Wayland will notify different values to each window/widget. For example, you have an app and it has two windows: one is laptop screen (x2 scale) and the other is external screen (x1.5 scale), their windows/widgets get `2` and `1.5` accordingly. But if you call `qApp->devicePixelRatio()`, it gives a single value (maybe `2`, maybe `1.5`).

tsujan: As I said, widget/window doesn't have meaning on the level of icon engines.

To see that widget/window scenario can't be true, it's enough to imagine that half of a single icon is on one screen and the other half on another screen.

guoyunhe: It is not a usual case to put a window across screens. But having different windows in different screens is very common. If users will complain about it, then it has some meanings.

[Qt doc](https://doc.qt.io/qt-5/qguiapplication.html#devicePixelRatio):

> Returns the **highest** screen device pixel ratio found on the system. This is the ratio between physical pixels and device-independent pixels.
>
> Use this function **only** when you don't know which window you are targeting. If you do know the target window, use `QWindow::devicePixelRatio()` instead.

tsujan: For the third time: `QWindow` is meaningless here.

Actually, I accept your point that `qApp->devicePixelRatio()` shouldn't be used here. If you have a practical suggestion, I'll be glad to hear it but please forget about `QWidget` and `QWindow`.

I'll return to this soon -- for now, I'm busy with a different case of unsharp icons, for which I'm making a patch...

tsujan: OK, back to this...

@guoyunhe The removal of `qApp` is possible in the following way:

```c++
return svgIcon.pixmap(size / svgIcon.pixmap(size, mode, state).devicePixelRatio(), mode, state);
```

The result is the same. But it's ugly because: (1) an extra pixmap is needed; and (2) it's still not obvious why SVG has this behavior.

tsujan:
Got rid of `qApp` and `devicePixelRatio` altogether by scaling the pixmap that the SVG icon engine gives. The final result is the same: sharp SVG icons with non-integer and integer scale factors alike.

tsujan:
This has fixed the old issue of unsharp SVG icons with scale factors here. Any objection to merging?

luis-pereira:
ship it

PR #211, opened 2020-06-24
Move member initialization to the initialize list

luis-pereira:
When an object of a class is created, the constructors of all member
variables are called consecutively in the order the variables are declared,
even if we don't explicitly write them to the initialization list.


tsujan:
GTM

PR #214, opened 2020-07-03
Workaround for the bug of Qt 5.15 in `QMouseEvent::buttons()` with `QMenu`

tsujan:
Closes https://github.com/lxqt/lxqt/issues/1821

Although this workaround works fine, I'm not sure if we should merge it. @luis-pereira, shouldn't we wait for a fix in Qt instead (even if it doesn't come soon)?

### DIFF
--- a/src/qtxdg/xdgmenuwidget.cpp
+++ b/src/qtxdg/xdgmenuwidget.cpp
@@ -158,7 +158,13 @@ bool XdgMenuWidget::event(QEvent* event)
 
 void XdgMenuWidgetPrivate::mouseMoveEvent(QMouseEvent *event)
 {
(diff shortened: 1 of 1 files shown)


luis-pereira:
@tsujan It works just fine.
Yes, Qt developers should fix that. It can wait, unless we start to receive a "lot's" of bug reports.

tsujan:
OK. So, let's keep it here without merging for a while. We could decide what to do if the bug isn't fixed when LXQt 0.16 is going to be released.

`QGuiApplication::mouseButtons()` seems reliable here because it's updated after a mouse button is pressed or released.

yan12125:
Closing as the issue has been fixed with Qt 5.15.1

PR #217, opened 2020-07-05
Fixed two memory leaks

tsujan:


src/qtxdg/xdgmimeappsglibbackend.cpp
```diff
@@ -42,6 +42,9 @@ static QList<XdgDesktopFile *> GAppInfoGListToXdgDesktopQList(GList *list)
                 if (df->load(file) && df->isValid()) {
                     dl.append(df);
                 }
+                else {
```
luis-pereira, line 45: Coding style not followed.

tsujan: > Coding style not followed.

`} else {` or something else? I did my best to follow the style ;)

luis-pereira: ```} else {```

there isn't any "example" in the all file.
Then merge it.

tsujan: Will change and merge it. Thanks!

PR #219, opened 2020-07-06
Fixed setting of default apps with GLib backend

tsujan:
In LXQt, although the GLib backend reads `~/.config/lxqt-mimeapps.list`, it writes to `~/.config/mimeapps.list` on setting default apps. Since we want to set default apps only for LXQt, the patch directly writes them to `lxqt-mimeapps.list` — after adding the needed associations to `mimeapps.list` (GLib always does the latter).

The patch considers `XDG_CURRENT_DESKTOP` and works outside LXQt too. In other words, it isn't limited to `lxqt-mimeapps.list`.

Fixes https://github.com/lxqt/libqtxdg/issues/218

### DIFF
--- a/src/qtxdg/xdgmimeappsglibbackend.cpp
+++ b/src/qtxdg/xdgmimeappsglibbackend.cpp
@@ -23,6 +23,7 @@
 
 #include "qtxdglogging.h"
 #include "xdgdesktopfile.h"
+#include "xdgdirs.h"
 
 #include <gio/gio.h>
 #include <gio/gdesktopappinfo.h>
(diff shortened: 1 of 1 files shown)


tsujan:
> If we are going this way we must take XDG_CURRENT_DESKTOP into account. libqtxdg isn't LXQt specific 

I have no problem with that.

> Does this rationale also applies to Added Associations and Removed Associations ?
No, because they only use `mimeapps.list`.

> It was a long long day, will continue tomorrow. Have to rest now.

Today was like 3 days to me ;) With these patches, I made default apps work at last. The only remaining challenge is finding a reliable way of restoring the previous settings. Honestly, `SettingsCache::loadToSettings` is very inefficient (https://github.com/lxqt/liblxqt/issues/262 is only one part of the problem). Will find a way tomorrow; should sleep too.

tsujan:
Updated the patch to consider `XDG_CURRENT_DESKTOP` and use `XdgDirs::configHome()`.

luis-pereira:
> No, because they only use `mimeapps.list`.

You mean the GLib implementation. I was referring to the standard.

> > It was a long long day, will continue tomorrow. Have to rest now.
> 
> Today was like 3 days to me ;) With these patches, I made default apps work at last. The only remaining challenge is finding a reliable way of restoring the previous settings. Honestly, `SettingsCache::loadToSettings` is very inefficient ([lxqt/liblxqt#262](https://github.com/lxqt/liblxqt/issues/262) is only one part of the problem). Will find a way tomorrow; should sleep too.

`SettingsCache::loadToSettings()` is very old. I remember that when I first downloaded Razor-qt source, it was already there.

tsujan:
> You mean the GLib implementation. I was referring to the standard.

I just supposed that GLib follows standards, knowing that Gnome devs have had a big part in implementing them.

GLib does this: It gives priority to the DE-specific `mimeapps.list` when reading default apps but ignores it when setting default apps. That isn't good even if it's a standard -- IMO, it's self-contradictory and a bug.

Apart from it, when GLib sets a default app, it adds the needed associations to the DE-independent `mimeapps.list`. I think that's reasonable because, although those associations may already exist, it's better to be safe than sorry.

PR #201, opened 2020-02-09
Fix chronic memory leak

stream009:
You can confirm memory leak by
1. launch lxqt-panel (with main menu plugin) or lxqt-runner
2. check memory usage with ps or top
2. trigger menu rebuild with
    `$ sudo touch /usr/share/applications` (or equivalent depend on distro)
3. repeat 2 for some time with few seconds interval
4. observe RSS + Swap increase

PR #221, opened 2020-07-20
processor: Fix memory leak

palinek:
supersedes #201


tsujan:
Nice. As I mentioned [elsewhere](https://github.com/lxqt/libqtxdg/pull/201#issuecomment-660826977), I tested it with valgrind and also by using it. Its logic is sound too. GTM.

PR #222, opened 2020-07-24
Xdg default apps

luis-pereira:


tsujan:
Very good. After this, `lxqt-config-file-associations` can be simplified and enhanced a little.

With the default browser being recognized, isn't it better to add the default email client and file manager to `XdgDefaultApps` too?

luis-pereira:
> With the default browser being recognized, isn't it better to add the default email client and file manager to `XdgDefaultApps` too?

That's the plan. Dd not had the time, tough.

tsujan:
> That's the plan. Dd not had the time, tough.

There's no hurry. When your work is finished, I'll modify the code of `lxqt-config-file-associations`.

luis-pereira:
> There's no hurry. When your work is finished, I'll modify the code of `lxqt-config-file-associations`.

Sure. This is a draft still. Didn't mark it as such, tough.

luis-pereira:
@tsujan Done.

PR #227, opened 2020-10-07
XdgMimeApps: Fix a memory leak

luis-pereira:


src/qtxdg/xdgmimeapps.cpp
```diff
@@ -96,6 +96,7 @@ QList<XdgDesktopFile *> XdgMimeApps::categoryApps(const QString &category)
         if (!categories.isEmpty() && (categories.contains(cat) || categories.contains(QL1S("X-") + cat)))
             dl.append(df);
     }
+    qDeleteAll(apps);
```
tsujan, line 99: Doesn't this make `df` consist of dangling pointers? I think we may need to delete only those pointers that aren't included in `df`.

luis-pereira: You were faster. That was a debug branch to find the source of the leak. Pushed the wrong branch. Fixed now.

luis-pereira:
@tsujan ping.

tsujan:
Oh, it's very good.

tsujan (approved):
GTM

PR #231, opened 2020-10-14
Fix a FTBFS with Qt versions earlier than 5.14

luis-pereira:
Using conditional compilation to avoid warnings with Qt versions >=Qt5.14.
Fixes https://github.com/lxqt/libqtxdg/issues/230.


palinek (commented):
GTM

PR #238, opened 2020-12-29
Improve readability in container emptiness check

luis-pereira:
Use empty() instead of size() is easier to understand.
Also avoids an implicit cast.


tsujan:
GTM

PR #239, opened 2020-12-29
Port QStringRef to QStringView

luis-pereira:
Qt6 dropped QStringRef.
Note: QStringViews should be passed by value, not by reference-to-const.
Passing a reference-to-const, compiles and works, but is slower.


tsujan:
GTM

PR #241, opened 2021-01-12
Move to Qt5 signal/slot syntax

adjamhub:
Just one notable change. I had to set the public inheritance of XdgMenuPrivate from QObject. It doesn't work otherwise. Anyway I coudn't find one reason why it has to privately inherit.


palinek:
> Anyway I coudn't find one reason why it has to privately inherit.

It looks to me like a missing explicit keyword from the day one :)

tsujan (approved):
GTM

PR #242, opened 2021-02-02
Add XTerm as a runtime dependency

luis-pereira:
A quick mitigation to https://github.com/lxqt/lxqt/issues/1889.
Note to packagers: Add XTerm as a recommended dependency.


tsujan (approved):
GTM

PR #243, opened 2021-02-04
Don't shadow variables

luis-pereira:
A local variable shouldn't shadow a outer variable.


tsujan:
It had no problem, IMHO, but yes, avoiding it may prevent confusion for the code reader. GTM

palinek (approved):
GTM

luis-pereira:
The compiler keeps track of the scope. But the reader, might not. This is a readability issue, of course.

PR #244, opened 2021-02-08
xdgdesktopfile: Allow non-detach starting

palinek:
Allow runtime overriding of detachment of processes by setting the
QTXDG_START_DETACH_TRULY. Default value is true [1].
With this the application (using libqt5xdg) can correctly send
termination signal to child processes on the application exit.

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -57,6 +57,7 @@
 #include <QTextStream>
 #include <QUrl>
 #include <QtAlgorithms>
+#include <QCoreApplication>
 
 
 /**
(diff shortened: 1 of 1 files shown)


palinek:
With this we can help applications started by lxqt-session/panel/runner to quit gracefully.

Ref. e.g. -> https://github.com/pvanek/qlipper/issues/34#issuecomment-248535444

tsujan:
> With this we can help applications started by lxqt-session/panel/runner to quit gracefully.

It'll be very useful because apps may save some of their settings (window size, cursor position, etc.) when they exit.

luis-pereira (approved):
GTM.

PR #247, opened 2021-03-18
Use QSvgRenderer for SVG icons

tsujan:
`QSvgRenderer` is used with a cache for both ordinary and colorized SVG icons. In this way, LXQt's icon handling cannot be broken by intruding icon engines, which register themselves for "svg" (see https://github.com/lxqt/libqtxdg/issues/246). Moreover, it does not depend on a specific code structure of `qtsvg` or any other icon engine.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -53,6 +53,7 @@
 #include <QImageReader>
 #include <QXmlStreamReader>
 #include <QFileSystemWatcher>
+#include <QSvgRenderer>
 
 #include <private/qhexstring_p.h>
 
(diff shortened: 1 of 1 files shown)


src/xdgiconloader/xdgiconloader.cpp
```diff
@@ -787,121 +788,142 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
...
+            bgCol = pal.highlight().color().name();
+        }
+        else // normal or active
+        {
+            txtCol = pal.windowText().color().name();
+            bgCol = pal.window().color().name();
+        }
+        hCol = pal.highlight().color().name();
+    }
+    QString key = QLatin1String("lxqt_")
```
palinek, line 863: Just suggestion... wouldn't it be better to to use the same key structure for both `ScalableFollowsColorEntry` & `ScalableEntry`? Or be the one at least prefix of the extended?

palinek (commented):
Perfect... get rid of relying on private undocumented internals is a good way to go.

src/xdgiconloader/xdgiconloader.cpp
tsujan: The only difference is that, in `ScalableFollowsColorEntry`, we need three colors (not the whole palette) but we don't need them in `ScalableEntry`.

BTW, what do you think about increasing `QPixmapCache` limit? I didn't think that was needed but wanted to ask your opinion.

palinek: > but we don't need them in ScalableEntry.

Sure... but then wouldn't it be better to have the same fixed-width prefix... like:
```cpp
      QString key = QLatin1String("lxqt_")
                    % filename
                    % HexString<int>(mode)
                    % HexString<int>(state)
                    % HexString<int>(size.width())
                    % HexString<int>(size.height())
                    % txtCol % bgCol % hCol;
```

..and can't the concatenation of *Col allow a collision?

> what do you think about increasing QPixmapCache limit?

Never looked into that. But I we need to increase...then I would vote for allowing it to be configurable.

tsujan: > but then wouldn't it be better to have the same fixed-width prefix... like:

No objection from me if you prefer it. No collision will happen either. Will do it soon.

> But if we need to increase...then I would vote for allowing it to be configurable

I agree. Haven't found a real use case yet; just an idea.

tsujan: Done; made the cache keys similar to each other.

palinek (approved):
Code review + simple test. Working as expected.

PR #249, opened 2021-04-15
Added a nullity check for paint device

tsujan:
It may not be needed in practice but Qt does it in `qicon.cpp` → `QPixmapIconEngine::paint()`.


palinek (approved):
GTM

PR #251, opened 2021-04-15
Fix a QStringView related FTBFS

luis-pereira:
QStringView was introduced in Qt 5.10 but some methods were introduced in
later versions. QStringView::lastIndexOf() was introduced in Qt 5.14.
Fixes https://github.com/lxqt/libqtxdg/issues/250.


tsujan (approved):
Thank you very much!

It's difficult to know these things by reading Qt doc — to say nothing of mistakes in Qt doc.

PR #256, opened 2021-04-29
Desktop file start handle relative paths

luis-pereira:


util/qtxdg-desktop-file-start.cpp
```diff
@@ -50,20 +50,21 @@ int main(int argc, char *argv[])
...
     QStringList userArgs = parser.positionalArguments();
     const QString userFileName = userArgs.takeFirst();
-    const QFileInfo fileInfo(userFileName);
-    const QString canonicalFileName = fileInfo.canonicalFilePath();
 
-    if (!fileInfo.exists()) {
-        printErr(QString::fromLatin1("File %1 does not exist\n").arg(userFileName));
-        return EXIT_FAILURE;
+    const QFileInfo fileInfo(userFileName);
+    if (fileInfo.isAbsolute()) {
```
palinek, line 55: Just curious... why is the `isAbsolute()` check needed? Is there a bug in Qt which turns out the `exists()` on relative path to be unreliable?

luis-pereira: Just to quit as soon as possible.
There's no Qt bug that I know of.
A simple `exists()` at this level doesn't cut it. If the path is relative a search through `XDG_DATA_HOME` and `XDG_DATA_DIRS` must be performed. `XdgDesktopFile` does that.

tsujan (approved):
Its logic seems good to me.

PR #257, opened 2021-05-15
XdgMenuWidget: fix for sometimes dragging incorrect action

slidinghotdog:
When starting drag, it should get the action that was clicked instead of the action currently under the cursor when qt detects drag action.

Causes issues in `lxqt-panel/plugin-mainmenu` when dragging items to quicklaunch or desktop, causing the adjacent item to be dragged instead.


tsujan (approved):
Seems logical.

PR #261, opened 2021-06-03
Make sure `hicolor` is searched before dash fallbacks

tsujan:
Dash fallbacks can cause all sorts of troubles. Although the inherited themes were searched *before* them, the code first ignored `hicolor` among the inherited themes and then searched it *after* dash fallbacks. That's corrected by the current patch.

Fixes https://github.com/lxqt/libqtxdg/issues/260. That report is also an example of how annoying dash fallbacks could be if they aren't used *only* as the last resort.


tsujan:
OK, after more than one month and because (1) it's tested by me and @stefonarch and (2) since it fixes https://github.com/lxqt/libqtxdg/issues/226 (@flychenjun also confirmed that in https://github.com/lxqt/libqtxdg/pull/255 but, for some reason unknown to me, deleted his comment when he closed that PR), I merge it and take full responsibility. If someone finds that a disaster happens with it, it could be easily reversed but that seems impossible to me.

PR #262, opened 2021-07-29
Consider XDG_DATA_HOME when computing desktop entry id

luis-pereira:
The standard doesn't refer to XDG_DATA_HOME. It should be considered, allowing
the id of an desktop entry located in XDG_DATA_HOME to be properly computed.
XDG_DATA_HOME is taken into consideration when searching for a desktop entry,
it make no sense not to consider it for computing id.

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -1230,9 +1230,9 @@ QString XdgDesktopFile::id(const QString &fileName, bool checkFileExists)
     }
 
     QString id = f.absoluteFilePath();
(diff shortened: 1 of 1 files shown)


tsujan:
Just a question: Is there any system, on which `XDG_DATA_DIRS` doesn't include `XDG_DATA_HOME`?

luis-pereira:
Openbox.
And for that matter any environment that doesn't set `XDG_DATA_DIRS`. It will fallback to 

- `/usr/local/share`
- `/usr/share`

tsujan:
Thanks!

PR #263, opened 2021-07-29
qtxdg-mat: Improve file manager command line handling

luis-pereira:
Handle arguments given without options.


luis-pereira:
Can somebody review this ?

tsujan:
Its code seems good. I think it's safe to merge it.

PR #268, opened 2021-11-30
qtxdg-mat: Rework exit codes

luis-pereira:
An asked --help must return EXIT_SUCCESS.
Any option passed to an non existent module must return EXIT_FAILURE.

### DIFF
--- a/src/tools/mat/qtxdg-mat.cpp
+++ b/src/tools/mat/qtxdg-mat.cpp
@@ -95,9 +95,26 @@ int main(int argc, char *argv[])
     // Find out the positional arguments.
     parser.parse(QCoreApplication::arguments());
     const QStringList args = parser.positionalArguments();
-    const QString command = args.isEmpty() ? QString() : args.first();
-    bool cmdFound = false;
+    if (args.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/tools/mat/qtxdg-mat.cpp
```diff
@@ -95,9 +95,27 @@ int main(int argc, char *argv[])
     // Find out the positional arguments.
     parser.parse(QCoreApplication::arguments());
     const QStringList args = parser.positionalArguments();
-    const QString command = args.isEmpty() ? QString() : args.first();
-    bool cmdFound = false;
+    parser.parse(QCoreApplication::arguments());
```
tsujan, line 98: Is there a reason to repeat it here? `parser.parse()` was called above, before a `const` method.

luis-pereira: No, just a leftover typo. Fixed.
Thanks.

luis-pereira: No, just a leftover typo. Fixed.
Thanks.

tsujan (approved):
GTM

PR #269, opened 2021-12-01
qtxdg-mat: Handle '--help-all' option

luis-pereira:
Wasn't being handled.
Just a documentation error, or a shortcoming in the API.
Upstream Qt bug description: https://bugreports.qt.io/browse/QTBUG-92567


tsujan (approved):
GTM (by looking at the changes).

PR #270, opened 2022-01-25
Consider single-quoted strings to be literals in Exec keys

tsujan:
According to Desktop Entry Specification, we should *not* do so but, apparently, most DEs do it.

Closes https://github.com/lxqt/lxqt-panel/issues/1705

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -1212,7 +1212,17 @@ QStringList XdgDesktopFile::expandExecString(const QStringList& urls) const
         }
 
         // ----------------------------------------------------------
-        result << expandEnvVariables(token);
+        if (token.startsWith(QLatin1Char('\'')) && token.endsWith(QLatin1Char('\'')))
+        {
+            // Consider 'XXX' to be a string literal.
+            // WARNING: This is beyond Desktop Entry Specification but,
+            // apparently, most DEs do it.
+            result << token.chopped(1).remove(0, 1);
+        }
+        else
+        {
+            result << expandEnvVariables(token);
+        }
     }
 
     return result;
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -1005,7 +1005,9 @@ static QStringList parseCombinedArgString(const QString &program)
     QStringList args;
     QString tmp;
     int quoteCount = 0;
+    int singleQuoteCount = 0;
     bool inQuote = false;
+    bool inSingleQuote = false;
 
     // handle quoting. tokens can be surrounded by double quotes
     // "hello world". three consecutive double quotes represent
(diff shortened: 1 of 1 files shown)


src/qtxdg/xdgdesktopfile.cpp
```diff
@@ -1212,7 +1212,17 @@ QStringList XdgDesktopFile::expandExecString(const QStringList& urls) const
         }
 
         // ----------------------------------------------------------
-        result << expandEnvVariables(token);
+        if (token.startsWith(QLatin1Char('\'')) && token.endsWith(QLatin1Char('\'')))
```
palinek, line 1215: This doesn't seem correct for me. 
1. consider situation, when the `'` is part of the file name -> it will be stripped wrongly
2. consider situation, when the `%f` would be surrounded by `'` in .desktop (for any reason) -> the `%f` won't be correctly expanded
3. even the `static QStringList parseCombinedArgString(const QString &program)` is wrong and does not consider escapes defined by spec

palinek (changes requested):
I believe the whole parameter parsing would need more love.

src/qtxdg/xdgdesktopfile.cpp
tsujan: Thanks for the review!

> consider situation, when the ' is part of the file name -> it will be stripped wrongly

It won't because the condition isn't satisfied. A specific example would be very appreciated.

> consider situation, when the %f would be surrounded by '

There's no change of behavior before and after the patch as far as I understand your point. A specific example?

tsujan:
After doing more tests, I agree that the parameter parsing is wrong *regardless of the patch*. The patch doesn't affect that — what was wrong remains wrong with it, under unusual but valid circumstances.

palinek:
>>consider situation, when the ' is part of the file name -> it will be stripped wrongly

>It won't because the condition isn't satisfied. A specific example would be very appreciated.

Any file, which name is (un)intentioanly enclosed `'` `'i_am_the_file_with_staring_and_ending_with_single_quote'`. The spec by itself doesn't state anything about absolute or relative file paths.

>>consider situation, when the %f would be surrounded by '

> There's no change of behavior before and after the patch as far as I understand your point. A specific example?

You are right. The same would happen even w/o your patch. But with the patch, the app won't get `'%f'`, but `%f`.

tsujan:
Please read what comes after "EDIT:" in my previous comment (I added it a few seconds before your last comment).

It seems to me that the problem may be much more complex that what's considered in `xdgdesktopfile.cpp`. The more I read it, the more counterexamples come to my mind. How it could be solved, I have no clear idea yet.

tsujan:
My conclusion:

`xdgdesktopfile.cpp` works under "ordinary" circumstances. For it to work under all valid circumstances, either it should be rewritten or some fixes should be applied to it. Only then, we could consider single quotes in a self-consistent way.

One "easy" way out of this situation could be rewriting `xdgdesktopfile.cpp` by using `GLib` (which is already used by `libqtxdg`).

Closing...

PR #273, opened 2022-02-08
Supported set use the dash fallback for icon name on XdgIconLoader

zccrs:
Add XdgIconLoader::dashFallback XdgIconLoader::setDashFallback.
Allows the caller to decide whether to use "dash fallback" behavior,
when finding some type of icon, often expects to do dash fallback
operation, such as
https://specifications.freeDesktop.org/icon-naming-spec/icon-naming-spec
-latest.html icon of the MimeTypes class mentioned.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -569,7 +569,7 @@ QThemeIconInfo XdgIconLoader::loadIcon(const QString &name) const
     const QString theme_name = QIconLoader::instance()->themeName();
     if (!theme_name.isEmpty()) {
         QStringList visited;
-        auto info = findIconHelper(theme_name, name, visited, true);
+        auto info = findIconHelper(theme_name, name, visited, m_dashFallback);
         if (info.entries.isEmpty()) {
             const auto unthemedInfo = unthemedFallback(name, QIcon::themeSearchPaths());
             if (unthemedInfo.entries.isEmpty()) {
--- a/src/xdgiconloader/xdgiconloader_p.h
+++ b/src/xdgiconloader/xdgiconloader_p.h
@@ -140,6 +140,8 @@ public:
      */
     inline bool followColorScheme() const { return m_followColorScheme; }
     void setFollowColorScheme(bool enable);
+    inline bool dashFallback() const { return m_dashFallback; }
+    void setDashFallback(bool fallback) { m_dashFallback = fallback; }
 
     XdgIconTheme theme() { return themeList.value(QIconLoader::instance()->themeName()); }
     static XdgIconLoader *instance();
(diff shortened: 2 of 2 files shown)


tsujan:
Would you please tell us where `freedesktop.org` says that dash fallback can be optional?

zccrs:
> Would you please tell us where `freedesktop.org` says that dash fallback can be optional?

The dash fallback is not optional in the `freedesktop.org`, but it should exclude the `MimeTypes` contexts. ref from https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec:

````
The dash “-” character is used to separate levels of specificity in icon names, for all contexts other than MimeTypes. For instance, we use “input-mouse” as the generic item for all mouse devices, and we use “input-mouse-usb” for a USB mouse device. However, if the more specific item does not exist in the current theme, and does exist in a parent theme, the generic icon from the current theme is preferred, in order to keep consistent style.
````

tsujan:
How is this PR related to what you quoted?

tsujan:
Dash fallbacks are already considered, in a way that's better for UX (like in KDE). See the comment that starts with https://github.com/lxqt/libqtxdg/blob/add94435fbe699610ab0eb61053945e1f7ce462d/src/xdgiconloader/xdgiconloader.cpp#L354 It includes what you quoted.

tsujan:
If your emphasis is on MimeTypes, this PR doesn't do anything either.

However, if dash fallbacks were skipped with MimeTypes, many icons might be replaced by "unknown". `freedesktop.org` isn't practical in this case. It's followed as far as helpful.

There are other parts of `freedesktop.org` that even Gnome doesn't follow word by word (e.g., desktop entry spec).

zccrs:
> If your emphasis is on MimeTypes, this PR doesn't do anything either.
> 
> However, if dash fallbacks were skipped with MimeTypes, many icons might be replaced by "unknown". `freedesktop.org` isn't practical in this case. It's followed as far as helpful.
> 
> There are other parts of `freedesktop.org` that even Gnome doesn't follow word by word (e.g., desktop entry spec).

I admit, the problem I'm having has nothing to do with Mimetypes. As you said, the freedesktop.org is not always correct in all cases, and some icons do not want to use the dash fallback behavior (eg, the icons of PWA applications created by chrome all starts with "chrome-", if a theme contains an icon named "chrome", then all PWA applications will use the same icon(The "chrome"), not the PWA application's own icon), so this commit can allow the caller to decide whether it should use the dash fallback, so libqtxdgicon That is, there is no intentional violation of the freedesktop.org, and it also allows the caller to use the dash fallback behavior flexibly.

zccrs:
I see the "https://github.com/lxqt/libqtxdg/pull/261" MR，It's can fix the chrome problom.

PR #275, opened 2022-03-03
Iconloader scaling enhancemets

palinek:
While debugging the lxqt/libfm-qt#788 I dived "deeply" into some aspects of the iconloader logic and here are probable enhancements.


palinek:
@tsujan Can you, please, have a look into this commits?

palinek:
> Anyway, this patch causes a crash with a scale factor:

Did you rebuild the `lxqt-qtplugin` also?

tsujan:
Oh, you're right; I forgot that. But, still, I think it's important to stick with how Qt does it, in order to guarantee that our code is as standard as possible.

palinek:
> I think it's important to stick with how Qt does it

If we don't fix spotted bugs/problems, then there is no point for us to have our own implementation.

tsujan:
> If we don't fix spotted bugs/problems, then there is no point for us to have our own implementation.

Yes, but I'm not sure to which bug you refer. https://github.com/lxqt/libfm-qt/pull/792 isn't a bug in `libqtxdg` and the situation can become very confusing because of problems that are really elsewhere (like in the Breeze icon set: https://github.com/lxqt/libfm-qt/pull/792#issuecomment-1057986354).

palinek:
> I'm not sure to which bug you refer

There are 3 commits here:
* iconloader: Build/use map for finding entries
** [current] the resolved entries for one particular `XdgIconLoaderEngine` object are crawled and tried to find the best match every time some information is requested
** [new] by adding a map the best suited entry for particular size&scale is searched only the first time; each next time a "cached pointer" is used (which I believe is more performant)
* iconloader: Handle non-integer scaling more precise
** [current] Qt (or any caller) calls the `QIconEngine` API requesting the <scaledSize> (<size> * <floating scale>); In `XdgIconLoaderEngine` for getting the most valid entry the it uses <requestedSize> (<scaledSize> / qCeil(<floating scale>)), which is obviously not correct as the original <size> != <requestedSize> for non-integer scales
** [new] consider the <floating scale> as is for gathering the <requestedSize>, but do round the <floating scale> only for evaluation of the scaled ("*@Sx") entries (also prefer the entries where scale == qCeil(<floating scale>))
* iconloader: Consider scale in actualSize() & pixmap()
** [current] in `actualSize()` & `pixmap()` for searching the most suitable entry the scale is not considered at all (i.e. it is (wrongly) assumed to be 1.0)
** [new] use application wide scale

Don't you think, it is worth addressing?

tsujan:
> Don't you think, it is worth addressing?

I'll need some time to read them.

Just one thing caught my eyes by a glance:
```c++
const qreal scale = qApp->devicePixelRatio();// Don't know which window to target
```
As far as I remember, we didn't have that elsewhere and I didn't find it in Qt's icon loader either. The problem is that a window of an app can be inside another screen with scaling. Therefore, `qApp->devicePixelRatio()` should be avoided, especially here. 

My general point is that such details could easily get out of hand after a while if we don't stick with Qt's way of doing things as far as possible. Moreover, if our code gets too much different from that of Qt, we'll have trouble in importing Qt's fixes (like https://github.com/lxqt/libqtxdg/pull/225).

palinek:
> My general point is that such details could easily get out of hand after a while if we don't stick with Qt's way of doing things as far as possible.

I believe all these commits can be ported to Qt straight away. But this probably won't be backported to Qt v5.15 (I don't know if Qt would accept this binary incompatible change).

palinek:
> Therefore, qApp->devicePixelRatio() should be avoided, especially here.

It is used in Qt for similar purposes also:
- https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L175
- https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L854
- https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L896
- https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L914
- https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L943

tsujan:
> It is used in Qt for similar purposes also:

It's the last resort and only where really needed. The pixel ratio of window, pixmap or paint device has priority over it.

Just one possible scenario among who knows how many?

I have an app with a `devicePixelRatio` of 2 on an HDPI screen. One of its windows opens in an ordinary screen. `QIcon::pixmap(QWindow *window,...)` calls `qt_effective_device_pixel_ratio()`, sets `devicePixelRatio` to 1, and then calls `d->engine->pixmap()`. With your change, it may get a wrong pixmap. It's only in `QIcon::pixmap(QWindow *window,...)` that we can know about a window; we shouldn't deal with the pixel ratio prematurely in `XdgIconLoaderEngine::pixmap()`.

palinek:
> One of its windows opens in an ordinary screen. QIcon::pixmap(QWindow *window,...) calls qt_effective_device_pixel_ratio(), sets devicePixelRatio to 1, and then calls d->engine->pixmap(). With your change, it may get a wrong pixmap. 

...and we can say the same for the former "HDPI windows" -> without the change the windows on all HDPI screens may get wrong pixmaps (because we assume the scale == 1).
There is no silver bullet for this, but I would say, that using the application wide scale will cover more situations correctly than assuming scale == 1.

tsujan:
> ...and we can say the same for the former "HDPI windows" -> without the change the windows on all HDPI screens may get wrong pixmaps (because we assume the scale == 1).

No. It's correctly handled by `QIcon::pixmap(QWindow *window,...`).

In old times, we relied on qApp's pixel ratio. Now things are different: qApp's pixel ratio shouldn't take priority and, especially, it shouldn't be considered prematurely.

One of our (or my) users warned me about this (don't remember where). My first reaction was like yours but then I saw its logic; made changes here (following Qt) as well as in Kvantum.

> There is no silver bullet for this

Following Qt is the safest way.

palinek:
> > > ...and we can say the same for the former "HDPI windows" -> without the change the windows on all HDPI screens may get wrong pixmaps (because we assume the scale == 1).

> >...and we can say the same for the former "HDPI windows" -> without the change the windows on all HDPI screens may get wrong pixmaps (because we assume the scale == 1).

> No. It's correctly handled by QIcon::pixmap(QWindow *window,...).

Then this exact answer is for you initial issue. We're not using/guessing the scale in PixmapEntry/ScalableEntry... but in the XdgIconLoaderEngine. Here I don't consider the usage pre-mature.

tsujan:
Qt6 has gone further and has removed the scaling from `QIconEngine::virtual_hook()` too. Now the caller is responsible for providing `QIcon::pixmap(const QSize &size, qreal devicePixelRatio, Mode mode, State state)` with the correct dpr.

palinek:
> Qt6 has gone further and has removed the scaling from QIconEngine::virtual_hook() too.

It dropped the whole `virtual_hook` concept, but this doesn't make any difference to our discussed topic here.

palinek:
>  and QIcon::pixmap(QWindow *window,...) has been deprecated. That's a big step further away from consulting qApp's dpr at the level of QIconLoaderEngine.

But there are still the `QIcon::pixmap(size...)` overloads, which just uses the `QIcon::pixmap(size, -1, ...)` and this handles the scaling:
```cpp
  QPixmap QIcon::pixmap(const QSize &size, qreal devicePixelRatio, Mode mode, State state) const
  {
      if (!d)
          return QPixmap();
  
      // Use the global devicePixelRatio if the caller does not know the target dpr
      if (devicePixelRatio == -1)
          devicePixelRatio = qApp->devicePixelRatio();
  
      // Handle the simple normal-dpi case
      if (!(devicePixelRatio > 1.0)) {
          QPixmap pixmap = d->engine->pixmap(size, mode, state);
          pixmap.setDevicePixelRatio(1.0);
          return pixmap;
      }
  
      // Try get a pixmap that is big enough to be displayed at device pixel resolution.
      QPixmap pixmap = d->engine->scaledPixmap(size * devicePixelRatio, mode, state, devicePixelRatio);
      pixmap.setDevicePixelRatio(d->pixmapDevicePixelRatio(devicePixelRatio, size, pixmap.size()));
      return pixmap;
  }
```
But reading the code of Qt (5.15 & 6.dev) and how the `QIcon` calls `engine->pixmap()`... yes you're right the `QIconEngine::pixmap(..)` should no use scaling.

But the `QIconEngine::actualSize(..)` is not aware of the scaling at all. So I still think, we should check tha app wide scale there.

tsujan:
> But the QIconEngine::actualSize(..) is not aware of the scaling at all.

In Qt5, because of the way `QIcon::actualSize(QWindow *window,...)` deals with it, `QIconEngine::actualSize()` shouldn't depend on qApp's dpr.


In Qt6, `QIcon::actualSize(QWindow *window,...)` is deprecated. Instead, `QIcon::actualSize(size)` takes qApp's dpr into account (and not that of window — there should be a good reason for it) and, again, because of that, `QIconEngine::actualSize()` shouldn't depend on qApp's dpr.

PR #279, opened 2022-04-03
Fixed config file name with empty `XDG_CURRENT_DESKTOP`

tsujan:
Previously, it started with dash.

See https://github.com/lxqt/libfm-qt/pull/766 for a similar fix.


luis-pereira (approved):
GTM.

Issue #260, opened 2021-06-01
Status Notifier not showing correct icons having dash in their names


ahsand97:
##### Expected Behavior

All applications that use appindicator should display their icon correctly in the panel.

##### Current Behavior

Some applications don't display their appindicator icon, it just shows the generic **application-x-executable** icon, Remmina is one of them, if I install Remmina normally and run it, it doesn't show the appindicator icon but it does display the icon correctly with the package Remmina-appindicator which is a Remmina version compiled with its own libappindicator as far as I know.
I'm developing some apps in Python and Java with GTK3 and have tested on both languages libraries that creates appindicators and same thing happens, no icon showing just the generic **application-x-executable**.

**Remmina package:** (it shows the generic **application-x-executable** icon for the appindicator)
![image](https://user-images.githubusercontent.com/32344641/120336532-0e9eef00-c2b8-11eb-87a7-063edd2fd1e3.png)

**Remmina-appindicator package:** (it shows correctly its icon for the appindicator)
![image](https://user-images.githubusercontent.com/32344641/120336820-5756a800-c2b8-11eb-8136-b5c0c10c4f58.png)

**JappIndicator library to create AppIndicators in Java:** (I've tested this on XFCE and GNOME and both works perfectly)
![image](https://user-images.githubusercontent.com/32344641/120337053-95ec6280-c2b8-11eb-8ed9-edad9010663c.png)

**SystemTray library to create AppIndicators in Java:** (tested on both XFCE and GNOME too and it works)
![image](https://user-images.githubusercontent.com/32344641/120337402-e5329300-c2b8-11eb-9de5-7ca1aa238bd7.png)

The only log I see when I run `lxqt-panel` from console and run those apps is this (don't really know if packages are missing):
```
Systray started
()
Error on DBus request(org.kde.StatusNotifierItem-1207-2,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.IconThemePath was not found in object /StatusNotifierItem)
Error on DBus request(org.kde.StatusNotifierItem-1208-1,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.IconThemePath was not found in object /StatusNotifierItem)
[...]

Issue #226, opened 2020-08-21
Icons of newly installed apps may still not be correct


tsujan:
In spite of using `QFileSystemWatcher` in `xdgiconloader.cpp` for keeping track of updated gtk icon caches, it's still possible that icons of newly installed apps are the fallback "application" icon in the main menu of LXQt Panel. Of course, a panel restart fixes the issue.

I encountered the problem when installing gnome-chess: its icon doesn't exist in my local active icon set but, although the set inherits "hicolor" (it has `Inherits=breeze,oxygen,gnome,hicolor` in its index) and the app icon was installed in "hicolor" and the gtk icon cache was updated with installation, the "application" icon was shown in the main menu.

Strangely enough, after I uninstalled gnome-chess, restarted the panel and re-installed gnome-chess, its icon was correct. So, I think the problem may be about timing.

The same thing happened with another app too.

NOTE: This is an old issue that was supposed to be fixed by `QFileSystemWatcher`. Before using `QFileSystemWatcher`, it always happened; now, it happens less often.

tsujan:
The problem seems to fixed by https://github.com/lxqt/libqtxdg/pull/261. I tested by installing gnome-chess and its icon appeared correctly with https://github.com/lxqt/libqtxdg/pull/261 but more tests are needed.

flychenjun:
I installed some wine applications and the icons are refreshed in time，The problem seems to fixed by #261

tsujan:
Fixed in https://github.com/lxqt/libqtxdg/pull/261, among other things.

tsujan:
Reopening because it partially happened when I installed `variety` (to test a user's claim, which wasn't related to this): The generic icon was shown in Main Menu before restarting the panel but the correct icon was shown in pcmanfm-qt without restarting it. It's a progress but not enough.

tsujan:
More thoughts:

As for why, after recent changes, this issue happens with lxqt-panel and not with pcmanfm-qt, my first thought was that it might be because of `XdgIcon::fromTheme()` but replacing it with `QIcon::fromTheme()` in the panel's code didn't make a difference.

`libfm-qt` (on which pcmanfm-qt depends) uses `GIcon` to find icons of desktop entries. My wild guess is that `GIcon` is somehow compatible with GTK icon cache, while Qt may have a "sync problem" with it.

tsujan:
The problem may not be in the cache but in how Qt updates icons.

[...]

PR #282, opened 2022-07-06
Fixed can't get icon when installing wine app

Dami-star:
When cache->isValid() returns true, cache->reValid(true) will no longer
be executed, resulting in gtkCachesWatcher->addPath no longer adding
monitoring files, which cannot be monitored when icon-theme.cache changes.

Even with 3.3.1, this fix still solves the problem

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -416,7 +416,7 @@ QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
             // Try to reduce the amount of subDirs by looking in the GTK+ cache in order to save
             // a massive amount of file stat (especially if the icon is not there)
             auto cache = theme.m_gtkCaches.at(i);
-            if (cache->isValid() || cache->reValid(true)) {
+            if (cache->reValid(true) || cache->isValid()) {
                 const auto result = cache->lookup(iconNameFallback);
                 if (cache->isValid()) {
                     const QVector<QIconDirInfo> subDirsCopy = subDirs;


Dami-star:
I just added my problem reproduction environment in #260

tsujan (changes requested):
I'm afraid this defeats the purpose of having a cache.

Also, see https://github.com/lxqt/libqtxdg/pull/255#pullrequestreview-644265682 and https://github.com/lxqt/libqtxdg/issues/226

Dami-star:
> I'm afraid this defeats the purpose of having a cache.
> 
> Also, see [#255 (review)](https://github.com/lxqt/libqtxdg/pull/255#pullrequestreview-644265682) and #226
Hi tsujan, thank you for your reply, I have spent a few days debugging this problem, I also refer to #255, my latest is to repeatedly install and uninstall multiple wine programs, and there will be a failure to get the icon. Especially when repeatedly When uninstalling and then installing.

try adding debug info here to identify the problem:

    qDebug() << "QIconCacheGtkReader::reValid, path :" << gtkCachesWatcher->files()<<m_cacheFileInfo.absoluteFilePath());
    if (!gtkCachesWatcher->files().contains(m_cacheFileInfo.absoluteFilePath()))
        gtkCachesWatcher->addPath(m_cacheFileInfo.absoluteFilePath());

Dami-star:
Or try to roll back the commit of #261, test and reproduce the problem, and then just use #226 to solve all problems

tsujan:
> Or try to roll back the commit of https://github.com/lxqt/libqtxdg/pull/261,..

Contrary to some comments, https://github.com/lxqt/libqtxdg/pull/261 had nothing to do with this. It was about consistency in treating dash fallbacks.

Yes, this PR avoids the problem by not using the cache as it's supposed to be used, but in that case, the removal of the cache would be a cleaner "solution". Neither is a real solution (→ https://github.com/lxqt/libqtxdg/issues/226).

Dami-star:
> > Or try to roll back the commit of #261,..
> 
> Contrary to some comments, #261 had nothing to do with this. It was about consistency in treating dash fallbacks.
> 
> Yes, this PR avoids the problem by not using the cache as it's supposed to be used, but in that case, the removal of the cache would be a cleaner "solution". Neither is a real solution (→ #226).

Thank you, do you have any good suggestions for changes to delete the cache

tsujan:
> do you have any good suggestions for changes to delete the cache

I didn't mean deleting the cache but *ignoring* it. Just remove it from `XdgIconLoader::findIconHelper`.

But I emphasize again: Without the cache, each icon should be found in the directory hierarchies of icon sets, and that takes more CPU time. It's not disastrous — you might not notice a difference with your computer — but it isn't good for all.

tsujan:
> Whether it is possible to judge whether the cache needs to be updated....

As I said in https://github.com/lxqt/libqtxdg/issues/226, the current code should do it; it uses `QFileSystemWatcher` for watching the cache. But, for some reason unknown to us, *sometimes*, it fails with newly installed apps.

Please read https://github.com/lxqt/libqtxdg/issues/226 and see how I was fooled into thinking that it was fixed (I closed and reopened the report). The problem definitely exists, in spite of watching the cache.

Dami-star:
> > Whether it is possible to judge whether the cache needs to be updated....
> 
> As I said in #226, the current code should do it; it uses `QFileSystemWatcher` for watching the cache. But, for some reason unknown to us, _sometimes_, it fails with newly installed apps.
> 
> Please read #226 and see how I was fooled into thinking that it was fixed (I closed and reopened the report). The problem definitely exists, in spite of watching the cache.

Ok, I have a good environment to reproduce the problem here, I am going to continue to find the cause of the problem from the cache, and if you have a good solution, I can help verify it.

Dami-star:
> > I have a good environment to reproduce the problem here
> 
> Very good.
> 
> A short explanation:
> 
> When the GTK icon cache changes (= is updated after app's installation), our watcher makes `QIconCacheGtkReader::isValid` return `false` and invalidates the icon (see the c-tor of `QIconCacheGtkReader`). As a result, `QIconCacheGtkReader::reValid` is called by `XdgIconLoader::findIconHelper`, the validity of the new cache is determined, and the icon is searched in it if valid. This sequence seems logical.
> 
> Now, the problem seems to be that, in the last stage, the new cache _is_ valid but doesn't contain the icon. My guess is that (1) it's somehow related to how `gtk-update-icon-cache` does its job, and (2) our watcher acts too soon, when the cache is changed but the icon isn't included in it yet.

Yes, I also think there is no problem with the logic of QIconCacheGtkReader::reValid, but what I have encountered is that after all the wine applications are uninstalled and the wine application is installed again, I cannot receive this signal QFileSystemWatcher::fileChanged

Dami-star:
BTW, every time we install and uninstall the wine application, icon-theme.cache is deleted and then recreated.
Similar to rm icon-theme.cache && new icon-theme.cache

AR0x7E7:
Hi gents, after installing two identical machines with the one difference of using lubuntu 18.04 and lubuntu 22.04 it is very frustrating to see my wine apps having the correct icons in 18.04, but in 22.04 it appears I need to uninstall and reinstall them again :(

Really hope that you can find a solution soon, thanks.

stefonarch:
> Really hope that you can find a solution soon, thanks.

Lubuntu 18.04 was using LXDE, Lubuntu 22.04 is using LXQt. 
It's not that there is any solution to be found and surely not by us.

Dami-star:
@tsujan HI tsujan, Is it possible to consider monitoring the folder where icon-theme.cache is located instead of the icon-theme.cache file to solve this problem.

tsujan:
> Is it possible to consider monitoring the folder...

The folder changes twice (at least): after the icon is installed/removed, and after the cache is updated.. Provided that we skip the first change(s) by checking the cache's modification time, we might get the current result with more computation — but it may be worth a try.

If we don't skip the first change(s), it'll be like ignoring the cache and searching in the dir hierarchy — something we could have done from start if it was acceptable.

tsujan:
I tried it; It didn't work. The result was the same as watching the cache file.

tsujan:
> This is not good news

No, it isn't. We're in the dark as to why the problem happens.

tsujan:
FYI, I also experimented with delaying the icon invalidation. Even with a 1-second delay, some icons weren't found.

Dami-star:
> FYI, I also experimented with delaying the icon invalidation. Even with a 1-second delay, some icons weren't found.
I also tested this method, and this delay time basically does not work when installing or uninstalling multiple applications at the same time

Dami-star:
> @Dami-star, see [#226 (comment)](https://github.com/lxqt/libqtxdg/issues/226#issuecomment-1191457085)
> 
> Closing this because it isn't a real solution...

Ok, I will use this scheme as a temporary scheme first, if I have a better scheme, I will open this and resubmit or apply for a pr

PR #283, opened 2022-07-22
Workaround for random lack of icons in newly installed apps

tsujan:
Workaround for random lack of icons in newly installed apps

This patch is a workaround for a random problem whose cause should be outside `xdgiconloader/xdgiconloader.cpp` — `xdgiconloader.cpp` seems correct to me in its original form.

The source of the problem is unknown to me, but it makes newly installed apps have `application-x-executable` as their icons in Main Menu.

The patch is a compromise: it consults GTK icon cache if valid; otherwise, it'll revalidate it and search in the directory hierarchy if the icon isn't found in it — which is the whole point of the patch.

IMHO, a DE that can't show the icons of newly installed apps has a serious UX issue.

Closes https://github.com/lxqt/libqtxdg/issues/226

EDIT: Changed the description a litttle.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -414,11 +414,11 @@ QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
             QVector<QIconDirInfo> subDirs = theme.keyList();
 
             // Try to reduce the amount of subDirs by looking in the GTK+ cache in order to save
-            // a massive amount of file stat (especially if the icon is not there)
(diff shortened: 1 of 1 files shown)


tsujan:
FYI,

Here, the search in dir hierarchy is done on two occasions:
 1. When the GTK cache is invalid (as before); or
 2. When the GTK cache is valid but doesn't include the icon.

Remove 2, and the problem will reappear. Why? After testing several hypotheses to no avail, I don't have the slightest idea.

2 is the cost. Supposing that the cache will be updated soon, it should be temporary (because of `cache->reValid(true);`).

PR #284, opened 2022-07-23
Invalidate GTK icon cache as soon as icons are installed

tsujan:
Monitoring the GTK cache itself wasn't enough for showing real icons of newly installed apps (in Main Menu). Contrary to my expectations, the reason was that the cache was invalidated too late.

Here, the folder is monitored instead, and the cache is invalidated as soon as something happens inside the folder. It will be validated again when it's updated.

Closes https://github.com/lxqt/libqtxdg/issues/226

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -149,14 +149,12 @@ QIconCacheGtkReader::QIconCacheGtkReader(const QString &dirName)
     // Note: The cache file can be (IS) removed and newly created during the
     // cache update. But we hold open file descriptor for the "old" removed
     // file. So we need to watch the changes and reopen/remap the file.
-    QObject::connect(gtkCachesWatcher(), &QFileSystemWatcher::fileChanged, &m_file, [this](const QString & path)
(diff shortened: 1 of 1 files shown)


Dami-star:
I have verified that the multiple installs of the application so far are fine

tsujan:
@Dami-star 
Thank you very much for the tests!

The core idea of this patch is yours. Here I just went in the opposite direction of what I did before.

I also installed (and uninstalled) ~15 apps after applying the patch, and their icons were shown correctly, both in Main Menu and in pcmanfm-qt. Considering your results alongside mine, I think this is the solution we searched for :)

Dami-star:
> @Dami-star Thank you very much for the tests!
> 
> The core idea of this patch is yours. Here I just went in the opposite direction of what I did before.
> 
> I also installed (and uninstalled) ~15 apps after applying the patch, and their icons were shown correctly, both in Main Menu and in pcmanfm-qt. Considering your results alongside mine, I think this is the solution we searched for :)

Great + 1, I also agree with this solution, I have tested and verified it under different architectures, and no problems have been found so far.

tsujan:
Any objection to merging this?

PR #285, opened 2022-08-10
Follow `freedesktop.org` in adding dirs under `DefaultDirectoryDirs`

tsujan:
Quoting https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html: "The default locations that are earlier in the search path go later in the `<Menu>` so that they have priority".

### DIFF
--- a/src/qtxdg/xdgmenureader.cpp
+++ b/src/qtxdg/xdgmenureader.cpp
@@ -349,6 +349,9 @@ void XdgMenuReader::processDirectoryDirTag(QDomElement& element)
  <DirectoryDir> elements containing the default desktop dir locations
  ($XDG_DATA_DIRS/desktop-directories/).
 
+ The default locations that are earlier in the search path go later in the <Menu>
+ so that they have priority.
+
  menu-cache additional prepends $XDG_DATA_HOME/applications.
  ************************************************/
 void XdgMenuReader::processDefaultDirectoryDirsTag(QDomElement& element)
@@ -357,8 +360,9 @@ void XdgMenuReader::processDefaultDirectoryDirsTag(QDomElement& element)
     QStringList dirs = XdgDirs::dataDirs();
     dirs.prepend(XdgDirs::dataHome(false));
 
-    for (const QString &dir : qAsConst(dirs))
-        addDirTag(element, QLatin1String("DirectoryDir"), dir + QLatin1String("/desktop-directories/"));
+    int n = dirs.size();
+    for (int i = 0; i < n; ++i)
+        addDirTag(element, QLatin1String("DirectoryDir"), dirs.at(n - i - 1) + QLatin1String("/desktop-directories/"));
 }
 
 /************************************************


stefonarch (approved):
Works as expected, priority has now  `~/.local/share/desktop-directories`

tsujan:
Yes, it seems priorities are OK now. Here, a panel restart isn't needed when adding/removing files — like in pcmanfm-qt → Applications.

stefonarch:
> Here, a panel restart isn't needed when adding/removing files

Yes, it's on the fly, the menu is generated at opening it seems to me. But I had to rebuild lxqt-panel against the branch.

src/qtxdg/xdgmenureader.cpp
```diff
@@ -357,8 +360,9 @@ void XdgMenuReader::processDefaultDirectoryDirsTag(QDomElement& element)
     QStringList dirs = XdgDirs::dataDirs();
     dirs.prepend(XdgDirs::dataHome(false));
 
-    for (const QString &dir : qAsConst(dirs))
-        addDirTag(element, QLatin1String("DirectoryDir"), dir + QLatin1String("/desktop-directories/"));
+    int n = dirs.size();
+    for (int i = 0; i < n; ++i)
```
palinek, line 364: I would rather use reverse iterators ...

tsujan: I don't think there would be a gain; would be?

palinek: I'm not sure. Just the computation of ```n - i - 1``` wouldn't be needed in each iteration.

tsujan: It's so cheap, perhaps the cheapest.

There are opposite theoretical or aesthetic opinions. I don't think we need to worry about them.

tsujan: An example, regarding `std::vector`: https://itecnote.com/tecnote/c-speed-accessing-a-stdvector-by-iterator-vs-by-operator-index/

palinek: > It's so cheap, perhaps the cheapest.

It may be the cheapest thing here, but it is completely unneeded and avoidable.

> An example, regarding std::vector: https://itecnote.com/tecnote/c-speed-accessing-a-stdvector-by-iterator-vs-by-operator-index/

I'm not arguing about iterators vs. `[]`, you're right, that in most cases that's just a personal preference.

Maybe I was bit misleading in the beginning with saying to rather use iterators. But using the iterators here would implicitly prevent to use the suboptimal index computation.

Anyway, even with the index loop I would suggest to use `for (int i = dirs.size() - 1; i >= 0; --i)`.

tsujan: We're only talking about different tastes here. The difference is zero in practice.

tsujan:
I asked for a review to make sure that this interpretation of `freedesktop.org` was correct (I assume the answer was positive). Feel free to make changes to the code style later. Merging...

PR #288, opened 2022-11-27
Fixed file association with symlink desktop entries

tsujan:
Previously, the (final) targets of symlink desktop entries were loaded and failed to meet the requirements of mime-type association. The patch solves the problem by loading the symlinks themselves.

I didn't find any side effect in the code or in practice. The only difference I've found is that, if a menu item corresponding to a symlink desktop entry is dragged and dropped, it'll be created as a symlink, not as its target — which isn't a problem and seems more natural to me.

Fixes https://github.com/lxqt/libqtxdg/issues/287 and closes https://github.com/lxqt/lxqt-config/issues/897

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -750,9 +750,8 @@ bool XdgDesktopFile::load(const QString& fileName)
 {
     d->clear();
     if (fileName.startsWith(QDir::separator())) { // absolute path
-        QFileInfo f(fileName);
-        if (f.exists())
-            d->mFileName = f.canonicalFilePath();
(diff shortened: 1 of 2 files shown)


tsujan:
Any objection to merging? It can be tested as described in https://github.com/lxqt/libqtxdg/issues/287 and https://github.com/lxqt/lxqt-config/issues/897.

stefonarch:
I had no issue, `*.csv` was assigned to Libreoffice calc, now  I changed to featherpad and couldn't switch back...

So 1) issue confirmed 2) patch is working fine.

tsujan:
Thanks for the confirmation.

The issue is very old. It has been hidden perhaps because symlinks are rare inside `/usr/share/applications`.

PR #292, opened 2023-03-07
Add support for translations defined via X-Ubuntu-Gettext-Domain

tsimonq2:
I'm submitting this upstream to be collaborative, but I can also understand why this could stay as a downstream patch. Not my decision, just figured I'd be nice and submit it. :)

The Lubuntu Global Team pointed out that some packages with Launchpad-based translations aren't being translated correctly in the LXQt menu. Further discussion can be found [here](https://discourse.lubuntu.me/t/problems-with-the-translations-of-desktop-launchers/4051).

In the interest of non-English users, we'll be carrying this in Lubuntu 23.04. If you guys have any code feedback, please let me know.

Thanks!

### DIFF
--- a/src/qtxdg/xdgdesktopfile.cpp
+++ b/src/qtxdg/xdgdesktopfile.cpp
@@ -920,7 +920,22 @@ QString XdgDesktopFile::localizedKey(const QString& key) const
 
 QVariant XdgDesktopFile::localizedValue(const QString& key, const QVariant& defaultValue) const
 {
-    return value(localizedKey(key), defaultValue);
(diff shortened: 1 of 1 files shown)


tsujan (changes requested):
Distro-specific codes should in no way appear here.

tsujan:
LXQt follows the "Desktop Entry Specification" of `freedesktop.org` as close as possible. This is how values are localized in a desktop entry: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s05.html

tsimonq2:
No worries, I agree this is non-standard and probably doesn't belong upstream.

PR #298, opened 2024-03-09
Qt5→Qt6

stefonarch:


README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-`libqtxdg` is a Qt 5 implementation of freedesktop.org XDG specifications.   
+`libqtxdg` is a Qt 6 implementation of freedesktop.org XDG specifications.
```
tsujan, line 5: What about "... is a Qt implementation of..."?

stefonarch: Preparing for Qt 7, why not.  It came to my mind that we could add instructions for building Qt5 (with `checkout <commit>` ?).

tsujan: The Qt version is obvious from the dependencies. The main info here is that libqtxdg is a *Qt implementation* of XDG specs.

tsujan:
Done in https://github.com/lxqt/libqtxdg/pull/301

PR #299, opened 2024-03-19
xdgdesktopfile: Consider XDG_CURRENT_DESKTOP env as ':' separated list

palinek:
Fixes lxqt/lxqt#2543


stefonarch:
GTM

```
$ echo $XDG_CURRENT_DESKTOP 
LXQt:kwin_wayland
```

stefonarch:
Maybe this could be modified to have a key in `.desktop` files to start them only under wayland in future?

stefonarch:
@palinek Now that we can have 
```
$ echo $XDG_CURRENT_DESKTOP 
LXQt:labwc:wlroots
```
couldn't we use a similar function as in https://github.com/lxqt/lxqt-session/pull/374/   which at the end after some seconds kills the compositor (together with  `lxqt-session` ...) if any found? Some of them has commands for exiting, e.g. `labwc -e` or `sway exit` ?

palinek:
But in wayland ecosystem, the lxqt-session isn't spawning WM/compositor, right? Why should it then be responsible for stopping it?

stefonarch:
> But in wayland ecosystem, the lxqt-session isn't spawning WM/compositor, right? Why should it then be responsible for stopping it?

Because `lxqt-leave` will close` lxqt-session` (which is started by the compositor), leaving  the compositor running which has to be closed by other means, so that leads to a doubled logout.

palinek:
The liblxqt's `LXQt::CustomProvider` ([src](https://github.com/lxqt/liblxqt/blob/master/lxqtpower/lxqtpowerproviders.cpp#L612-L696); highest priority in the provider's list) be used to achieve this -> define a `logoutCommand` in the config file. This command can be a simple script executing the desired commands, proposed by you, based on the env. var (or if nothing todo, just invoke the `logout` on lxqt-session via d-bus).

stefonarch:
Nice, didn't know that. For kwin and maybe later  als ino labwc an option is present: `--exit-with-session` which solves it  for those 2. Thanks!

stefonarch:
@palinek Unfortunately it doesn't work...

`power.conf` :
```
logoutCommand=exit_compositors
```
`/usr/local/bin/exit_compositors` :
```
#!/bin/bash

qdbus org.lxqt.session /LXQtSession org.lxqt.session.logout &&
sleep 2 &&

if [[ $XDG_CURRENT_DESKTOP == *"labwc"* ]];
	then
	labwc --exit
elif
	[[ $XDG_CURRENT_DESKTOP == *"sway"* ]];
	then
	sway exit
elif
	[[ $XDG_CURRENT_DESKTOP = *"wayfire"* ]]
	then
	killall wayfire
elif
	[[ $XDG_CURRENT_DESKTOP = *"Hyprland"* ]]
	then
	hyprctl dispatch exit
fi
```

I think the script exits together with `lxqt-session` and doesn't exit compositors therefore. It works fine if launched by terminal.

What works instead is  in `lxqt-session/lxqt-session/src/lxqtmodman.cpp`
```
    // terminate all possible children except WM
    mProcReaper.stop({mWmProcess->processId()});

    mWmProcess->terminate();
    if (mWmProcess->state() != QProcess::NotRunning && !mWmProcess->waitForFinished(2000))
    {
        qCWarning(SESSION) << "Window Manager won't terminate ... killing.";
        mWmProcess->kill();
    }
        // Under wayland kill the compositor

    if (QGuiApplication::platformName() == QLatin1String("wayland")) {
    QString program = QString::fromUtf8("exit_compositors");
    QStringList arguments;
    arguments << QStringLiteral("");
    QProcess::startDetached(program, arguments);
    }

    if (doExit)
        QCoreApplication::exit(0);
}

```

palinek:
It's because you have the `sleep` in your script there after issuing DBus logout to `lxqt-session`. `lxqt-session` terminates all children processes upon its finish. So it also terminates your `exit_compositors` script before it tries to shutdown the wayland compositor. To make it correctly you need to spawn the `exsit_compositors` logic as child of process 1, e.g. by using `at`, creating a systemd unit, etc.

stefonarch:
I added sleep only in a second attempt because what I saw before was that it didn't terminate child processes correctly but immediately executed the command to exit compositor.
Will retry.

PR #304, opened 2024-04-30
Added `plasma-applications.menu` to fallback menu files

tsujan:
Recently, `applications.menu` (belonging to `kservice5`) was removed and `plasma-applications.menu` (belonging to `plasma-workspace`) was added. Hence adding the latter to the fallback list.


luis-pereira (approved):
GTM.

PR #307, opened 2024-09-12
Fix compilation with Qt ≥ 6.8

tsujan:
Closes https://github.com/lxqt/libqtxdg/issues/306

### DIFF
--- a/src/qtxdg/xdgmimetype.h
+++ b/src/qtxdg/xdgmimetype.h
@@ -72,12 +72,12 @@ public:
      */
     bool operator==(const XdgMimeType &other) const
     {
-        return QMimeType::operator==(other);
+        return name() == other.name();
     }
 
     inline bool operator!=(const XdgMimeType &other) const
     {
-        return !QMimeType::operator==(other);
+        return !operator==(other);
     }
 
     void swap(XdgMimeType &other) noexcept;
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -794,7 +794,6 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
         calculatedDpr = qMax(qreal(1.0), scale * ratio);
     }
 
-
     QString key = QLatin1String("$qt_theme_")
                   % HexString<quint64>(basePixmap.cacheKey())
                   % HexString<quint8>(mode)
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -757,7 +757,11 @@ QSize XdgIconLoaderEngine::actualSize(const QSize &size, QIcon::Mode mode,
 }
 
 // XXX: duplicated from qiconloader.cpp, because this symbol isn't exported :(
+#if (QT_VERSION >= QT_VERSION_CHECK(6,8,0))
+QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state, qreal scale)
+#else
 QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state)
+#endif
 {
     Q_UNUSED(state);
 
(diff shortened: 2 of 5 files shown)


n3rdopolis:
What else do I need to test?, this compiles for me now, and I am pretty sure the scaling still works

tsujan:
> What else do I need to test?,

Nothing comes to my mind. You've already tested it, and the PR would be impossible without your contribution.

Chiitoo:
Qt 6.8.0 was released yesterday, though masked from normal testing on Gentoo Linux at least due to known issues out there.

(Compiling it here now though.)

stefonarch:
Oops, I

> I hope that Qt 6.8 will be released after LXQt 2.1.0. What should we do otherwise? Should we make a point release 4.0.1?

Well, hope is dead, I was about to say that we could also release earlier with QT6.7.
Point release is the only way no? And maybe other issues show up to. Hoping that some issues are fixed in QT6.8.

tsujan:
> Point release is the only way no?

OK. If no issue is found, I'll merge it the day after tomorrow and make the point release 4.0.1.

@Chiitoo, thanks for checking it!

Chiitoo:
So far so good here, with this change applied.

Here is some info on the known regressions and such:

https://mail.kde.org/pipermail/distributions/2024-October/001536.html

stefonarch:
> If no issue is found, I'll merge it the day after tomorrow and make the point release 4.0.1.

Not better wait until it's in arch?

tsujan:
We have about one month to the next release. I don't think a point release would make sense, for example, 2 weeks before the next release; does it?

yan12125:
Qt maintainer at Arch Linux is quite active. It may not take too long before Qt 6.8 enters Arch stable repos. Also, libqtxdg should be rebuilt with Qt 6.8 before new Qt can enter stable repos, as libqtxdg uses private Qt APIs.

src/xdgiconloader/xdgiconloader.cpp
```diff
@@ -766,18 +770,45 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
...
+    QSize targetSize = size * scale;
+    if ((actualSize.width() == targetSize.width() && actualSize.height() <= targetSize.height()) ||
+        (actualSize.width() <= targetSize.width() && actualSize.height() == targetSize.height()))
+    {
+        // Correctly scaled for dpr, just having different aspect ratio
+        calculatedDpr = scale;
+    }
+    else
+    {
+        qreal ratio = 0.5 * (qreal(actualSize.width()) / qreal(targetSize.width()) +
```
yan12125, line 792: 0.5 seems a magic number. We're duplicating Qt codes, anyway https://github.com/qt/qtbase/blob/v6.8.0/src/gui/image/qicon.cpp#L157

yan12125 (commented):
Just a minor comment after a quick glance

src/qtxdg/xdgmimetype.h
```diff
@@ -72,12 +72,12 @@ class QTXDG_API XdgMimeType : public QMimeType {
      */
     bool operator==(const XdgMimeType &other) const
     {
-        return QMimeType::operator==(other);
+        return name() == other.name();
```
palinek, line 75: Why dropping the `QMimeType` logic? IMO we can be more safe for future Qt changes, if we continue to stick with Qt's implementation:
```cpp
return static_cast<const QMimeType &>(*this) == other;
```

src/xdgiconloader/xdgiconloader.cpp
```diff
@@ -766,18 +770,45 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
...
+    }
+
+    QString key = QLatin1String("$qt_theme_")
+                  % HexString<quint64>(basePixmap.cacheKey())
+                  % HexString<quint8>(mode)
+                  % HexString<quint64>(QGuiApplication::palette().cacheKey())
+                  % HexString<uint>(actualSize.width())
+                  % HexString<uint>(actualSize.height())
+                  % HexString<quint16>(qRound(calculatedDpr * 1000));
+#else
```
palinek, line 804: Why the change in types of the key and/or why only applicable for Qt >= 6.8?
For the key assembly I would suggest to not repeat everything, but just add conditionally the new key, e.g.:
```cpp
    QString key = QLatin1String("lxqt_")
                  % filename
                  % HexString<quint8>(mode)
                  % HexString<int>(state)
                  % HexString<uint>(size.width())
                  % HexString<uint>(size.height())
#if (QT_VERSION >= QT_VERSION_CHECK(6,8,0))
                  % HexString<quint16>(qRound(scale * 1000))
#endif
                  ;
```
(the change would be better visible, IMO; applicable also in next changes)

palinek (commented):
My 2c just a classic source code readability nitpicking comments. I didn't dive into the logic and scale/dpr computations :(

src/xdgiconloader/xdgiconloader.cpp
yan12125: Good point!

src/qtxdg/xdgmimetype.h
tsujan: > Why dropping the QMimeType logic?

Basic things have changed in Qt 6.8 (also, see https://github.com/lxqt/libqtxdg/issues/306#issuecomment-2348995325). Moreover, two mimetypes are equal if their names are.

src/xdgiconloader/xdgiconloader.cpp
tsujan: As the comment `// see QIconPrivate::pixmapDevicePixelRatio` suggests, it's just that function, imported here.

I also added the comment `// see QPixmapIconEngine::adjustSize` for an already existing calculation to show its origin.

src/xdgiconloader/xdgiconloader.cpp
tsujan: It's very simple: I didn't want to change *anything* for Qt < 6.8 here. I really prefer it this way because it reflects the changes.

palinek: > It's very simple: I didn't want to change anything for Qt < 6.8 here.

Why did you change the types? There is nothing about it in commit message.

tsujan:
Thanks for the code reviews! One of them was only about different preferences regarding the code style.

But the other one was about `QMimeType`. Since it's definitely safe to compare names, I suggest we use this way for now and, if needed, try other approaches after things settle down.


As for the logic of scaling, I've checked it several times, by considering the changes made in Qt 6.8 (the note at https://doc.qt.io/qt-6/qicon.html#pixmap-3 may give you an idea), and think it should be correct. Of course mistakes are possible, and I can't be 100% sure until I get Qt 6.8 here.

tsujan:
Released [4.0.1](https://github.com/lxqt/libqtxdg/releases/tag/4.0.1).

PR #311, opened 2024-12-10
Adapt to changes in private headers Qt v6.9

### LINKED ISSUE #310: FTBFS with Qt 6.9.0
libqtxdg FTBFS against the tip of qtbase upstream with the following error:

```
In file included from /root/libqtxdg/src/xdgiconloader/xdgiconloader.cpp:37:
/root/libqtxdg/src/xdgiconloader/xdgiconloader_p.h:64:8: error: cannot derive from ‘final’ base ‘ScalableEntry’ in derived type ‘ScalableFollowsColorEntry’
   64 | struct ScalableFollowsColorEntry : public ScalableEntry
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
```

This is due to https://code.qt.io/cgit/qt/qtbase.git/commit/?id=aa38df53b977d8c04123f8c0994d041d183a7971

At the top of the header file there is:
```
//
//  W A R N I N G
//  -------------
//
// This file is not part of the Qt API.  It exists purely as an
// implementation detail.  This header file may change from version to
// version without notice, or even be removed.
//
// We mean it.
//
```

I'm not exactly sure how to address this.

palinek:
fixes #310

### DIFF
--- a/src/xdgiconloader/xdgiconloader_p.h
+++ b/src/xdgiconloader/xdgiconloader_p.h
@@ -61,13 +61,14 @@
 
 class XdgIconLoader;
 
-struct ScalableFollowsColorEntry : public ScalableEntry
+struct ScalableFollowsColorEntry : public QIconLoaderEngineEntry
 {
 #if (QT_VERSION >= QT_VERSION_CHECK(6,8,0))
     QPixmap pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state, qreal scale) override;
 #else
     QPixmap pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state) override;
 #endif
+    QIcon svgIcon;
 };
 
 //class QIconLoaderEngine : public QIconEngine
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -439,7 +439,11 @@ QThemeIconInfo XdgIconLoader::findIconHelper(const QString &themeName,
                 } else if (gSupportsSvg) {
                     const QString svgPath = subDir + svgIconName;
                     if (QFile::exists(svgPath)) {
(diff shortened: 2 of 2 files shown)


tsujan (approved):
GTM.

We may need a point release before Qt 6.9 comes out.

tsujan:
Unfortunately, this breaks our icon engine.

This is a desktop file in pcmanfm-qt's thumbnail view, with a 128-px size, and with libqtxdg 4.1.0:


And the same file with this patch:


The desktop file has `Icon=emblem-music-symbolic`, which is a 16-px SVG icon belonging to Breeze. Without the patch, it's correctly replaced by `emblem-music` from my icon set (according to our old rule: "give priority to the set, and only if not found, go to the inherited sets"); with the patch, it's taken from Breeze, which is inherited.

tsujan:
Solved the size problem by replacing

https://github.com/lxqt/libqtxdg/blob/1605b7d5db3e8e2855e8a1dbd0399c69a256650f/src/xdgiconloader/xdgiconloader.cpp#L740-L741

with

```c++
        if (dir.type == QIconDirInfo::Scalable
            || dynamic_cast<ScalableEntry *>(entry)
            || dynamic_cast<ScalableFollowsColorEntry *>(entry)) {
            return size;
        }
```

But the path problem still exists.

src/xdgiconloader/xdgiconloader_p.h
```diff
@@ -61,13 +61,14 @@
...
 
-struct ScalableFollowsColorEntry : public ScalableEntry
+struct ScalableFollowsColorEntry : public QIconLoaderEngineEntry
 {
 #if (QT_VERSION >= QT_VERSION_CHECK(6,8,0))
     QPixmap pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state, qreal scale) override;
 #else
     QPixmap pixmap(const QSize &size, QIcon::Mode mode, QIcon::State state) override;
 #endif
+    QIcon svgIcon;
```
tsujan, line 71: Isn't it better to put the above changes inside `#if (QT_VERSION >= QT_VERSION_CHECK(6,8,0))`? We still support Qt 6.6.

tsujan:
I've used this for months and encountered no problem. Any objection to merging?

PR #313, opened 2025-06-17
cmake: fix build with Qt 6.10

Chiitoo:
The 'Qt6FooPrivate' targets have been split into separate CMake files
in Qt 6.9, and require a 'find_package(Qt6FooPrivate)' call starting
with Qt 6.10.

See also: https://bugreports.qt.io/browse/QTBUG-87776


luis-pereira:
@Chiitoo Thanks.

PR #317, opened 2025-09-27
Use locale-aware sorting for menu entries

City-busz:
This fixes the problem that entries starting with non-English characters were placed at the end of the menu.


luis-pereira:
Landed at https://github.com/lxqt/libqtxdg/commit/d57acb50198f31659883ea2af0d3c03107f02ed5
@City-busz Thanks.

PR #323, opened 2025-11-17
Use "new" string nomenclature and classes

luis-pereira:


luis-pereira:
Reviewed it again. The only mistake I found was https://github.com/lxqt/libqtxdg/pull/323/commits/146071be2a4f12a388589e14c6f55a26c017533a.

tsujan:
Yes, I noticed: a needle in the haystack. Ship it if you have a good feeling about it.

PR #325, opened 2025-11-24
Prefer std:: functions

isf63:
Used `rg -P "(qSin|qCos|qTan|qAsin|qAcos|qAtan|qAtan2|qExp|qLn|qPow|qSqrt|qFabs|qFloor|qCeil|qRound|qRound64|qNextPowerOfTwo|qRadiansToDegrees|qHypot|qFabs|qSort|qStableSort|qLowerBound|qUpperBound|qBinaryFind|qCopy|qFill|qCount|qFind|qEqual|qShuffle|qLess|qGreater|qSwap|qMin|qMax|qAbs|qSwap)"`.

This didn't need to be done particularly, but they were all in one file...

The main component using the Qt functions _a lot_ is qtermwidget.

### DIFF
--- a/src/xdgiconloader/xdgiconloader.cpp
+++ b/src/xdgiconloader/xdgiconloader.cpp
@@ -811,7 +811,7 @@ QPixmap PixmapEntry::pixmap(const QSize &size, QIcon::Mode mode, QIcon::State st
                   % HexString<quint64>(QGuiApplication::palette().cacheKey())
                   % HexString<uint>(actualSize.width())
                   % HexString<uint>(actualSize.height())
-                  % HexString<quint16>(std::round(calculatedDpr * 1000));
+                  % HexString<quint16>(qRound(calculatedDpr * 1000));
 #else
     QString key = "$qt_theme_"_L1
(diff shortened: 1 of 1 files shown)


luis-pereira (changes requested):
qRound() vs std::round()
	qRound() aprox 9x times faster
	
qround() vs std::lround()
	qRound aprox 4x times faster
	
qCeil() vs static_cast<int>(std::ceil))
	Equivalent performance
	
qAbs() vs std::abs()
	Equivalent performance with O2
	qAbs() performs a assert range check in debug mode. Edge case, probably not needed. 
	
qMin() vs std::min()
	Equivalent performance.
	
I'm Ok with all except `std::round()`. Even `std::lround()` doesn't match the `qRound()` performance. `qRound` algorithm is good enough here.

luis-pereira:
It's Ok. I would prefer it in a single commit.
@isf63 Can you join them ?

isf63:
I will go ahead and squash and merge, if that's what you meant. I am known to make Git mistakes, so that should be avoided.

PR #328, opened 2026-01-17
Fix NotShowIn handling

City-busz:
This fixes the problem that setting `NotShowIn=LXQt` does not work.

This happened because the panel passes "X-LXQT" and "LXQt" values as environments, and these environments were checked separately. But to honor the `NotShowIn` value, it needs to be checked at once.


stefonarch:
Tested, works fine. Closes https://github.com/lxqt/lxqt/issues/1743
@palinek ? @luis-pereira ?

palinek:
Looks good.

tsujan:
When the code is reviewed by @palinek and tested by @stefonarch, why not merging it? Please don't wait for me; the situation is beyond surreal where I live, and it includes Internet too.

tsujan:
Oh, @luis-pereira should also check it, I think.

luis-pereira:
GTM.

luis-pereira:
@City-busz Thanks.

